PR #4, opened 2015-10-31
Andre solution: Ruby

codedre:
comfort : 4.5
completeness : 5

I was not sure how to go about making the instance method but when I looked at the solution it was what I thought, but before I did not know if we had to make a tenant's instance variable to hold all the tenants because initially I made the assumption that renters would just be used as an incrementer which in a way it still is by using the array length property.


codedre:
Submitting my tunr as a sub for landlord

Spent my time really understanding the process and will work on landlord over the weekend.

Completeness 4
comfort 4

https://github.com/codedre/tunr

mattscilipoti:
:+1:

mattscilipoti:
You are correct.  No need to make tenants instance var if I have access to an apartment.  I can just query `apartment.tenants`.

PR #6, opened 2015-11-03
pz solution - ruby, sql, AR

pzhang87:
comfort: 4.5
completeness: 5

SQL and AR make sense to me.  ~~Still need a little clarification on attribute helper methods (?) for some stuff, so I went with a simpler example (I think)~~

update with fixed attribute helper methods

### DIFF
--- a/controllers/tenants_controller.rb
+++ b/controllers/tenants_controller.rb
@@ -1,0 +1,3 @@
+get '/tenants' do
+  "ur mom"
+end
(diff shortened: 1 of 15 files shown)


controllers/tenants_controller.rb
```diff
@@ -0,0 +1,3 @@
+get '/tenants' do
+  "ur mom"
```
nolds9, line 2: Dorothy Mantooth is a saint!
https://www.youtube.com/watch?v=MsQH7PFrI04

nolds9:
:+1:  Great job on Part 1! Please reopen for submission of part 2

PR #8, opened 2015-11-03
Landlord solution joe gz

joe-gz:
comfort: 3
completeness: 4

through the active record portion i am done. still working on the bonus


ebirving:
:+1:

PR #9, opened 2015-11-04
Kain: Landlord Wednesday & Friday HW

elisekain:
Friday
Completeness: 5/5
Comfort Level: 4/5

Need to do this about 1000 more times, but basic CRUD functionality is working.

Wednesday
Completeness: 5/5
Comfort Level: 4/5 Schema & Ruby Classes. 2/5 Setting up Views/Templates in Sinatra.

I am having a hard time understanding how the views match up with our database and routes. 

Also, should the Gemfile.lock be going in our git repository? Or just the Gemfile...? Same question for Sass related files (sass-cache and styles.css.map)


public/styles.scss
```diff
@@ -0,0 +1,128 @@
...
+$hover: lighten($green, 10%);
+$white: #fff;
+
+/* Mixins */
+@mixin border-radius($radius) {
+  -webkit-border-radius: $radius;
+     -moz-border-radius: $radius;
+      -ms-border-radius: $radius;
+          border-radius: $radius;
+}
```
jsm13, line 14: Awesome

controllers/apartments.rb
```diff
@@ -0,0 +1,41 @@
...
+
+# Add apartment
+get "/apartments/new" do
+  erb :"apartments/new"
+end
+
+# Post new apartment form
+post '/apartments' do
+  apt = Apartment.create(address: params[:address], monthly_rent: params[:monthly_rent], sqft: params[:sqft], num_beds: params[:num_beds], num_baths: params[:num_baths])
+  redirect "/apartments/" + apt.id.to_s
```
jsm13, line 15: String interpolation could be handy here

jsm13:
:+1:  Super! Totally right about practice making everything make more sense and there will be plenty of practice. Has talking about MVC the last couple days helped with that relationships between the database, template, and routes? Very cool you got into Sass. Have you used Sass before?

controllers/apartments.rb
elisekain, line 15: Yes, yes it would be : )

elisekain:
MVC is much clearer now than when I initially wrote my comment. I'm demonstrating the extent of my current Sass knowledge...but want to make it a practice moving forward as I work on projects/homework to use it.

PR #10, opened 2015-11-04
Adds code I worked on with John Master

clarktsc39:
I've been staying after each day this week, and it's been helping a lot. I'm staying on top of my homework and learning a lot more. I'm going to continue to stay after from now on; wish I'd learned this 3 weeks ago -- oh, well...


amaseda:
Glad to hear staying after has helped. Definitely take advantage of that and office hours whenever you can!

db/seed.rb
```diff
@@ -0,0 +1,16 @@
...
+
+require "pg"
+require "active_record"
+require "pry"
+
+require_relative "connection"
+require_relative "../models/apartment"
+require_relative"../models/tenant"
+
+Apartment.create address ([
```
amaseda, line 11: This won't work with `address` between `.create` and the opening parentheses. The `.create` method looks something like this...

``` rb
Apartment.create( address: "123 Main St", monthly_rent: 5000 )
```

I like how you're passing in multiple hashes into the create method though!

views/Tenant.rb
```diff
@@ -0,0 +1,4 @@
+class Tenant
```
amaseda, line 1: A couple things...
1. Don't forge to close you class with `end`!
2. You should instantiate your instance variables inside of an initialize method. You also need to pass arguments into it. So something like this...

``` rb
class Tenant
  def initialize( name, age, gender )
    @name = name
    @age = age
    @gender = gender
  end
end
```

(Keep in mind, once you make your classes inherit from ActiveRecord, they won't need initialize methods anymore!)

amaseda:
Good start. Will take another look when you submit Landlord in full on Friday. Let me know if you have any questions in the meantime.

:+1:

PR #12, opened 2015-11-04
Everything was going fine...HW done

jnichols300:
Comfort 4/5 
Completeness 4.99/5

I spelled monthly wrong and messed up my database, then I dropped it and tried to follow the directions but I broke something and now I can't figure out why my database is messed up. Hopefully this is a quick easy fix I can do with an instructor's help.

### DIFF
--- a/db/app.rb
+++ b/db/app.rb
@@ -1,0 +1,18 @@
+require 'sinatra'
+require 'sinatra/reloader'
+
+get '/' do
+  erb:":apartments"
+end
+
+get '/apartments' do
+  erb:"apartments/1"
+end
+
+get '/apartments/new' do
+  erb:"apartments/new"
+end
+
+get '/apartments/1/tenants' do
+  erb:"apartments/1/tenants"
+end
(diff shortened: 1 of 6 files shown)


andrewsunglaekim:
:+1:  I'm sure its not a big deal, come see someone during office hours! Also, I dont see the views associated with with the routes you set up. And be careful in the naming of the routes. Other than that gj!

db/app.rb
```diff
@@ -0,0 +1,26 @@
...
+require_relative "models/apartment"
+require_relative "models/tenant"
+
+get '/' do
+  erb:":apartments"
+end
+
+get '/apartments' do
+  erb:"apartments/1"
+end
```
andrewsunglaekim, line 18: really the erb should be called `apartments/show` the route itself should be `get '/apartments/:is`. What do you do for an apartment that has an id of 2? Is a new route set up? The view should just be a template and it doesn't matter which apartment goes into that template it'll display the information in the same way.

db/app.rb
```diff
@@ -0,0 +1,26 @@
...
+
+get '/apartments/new' do
+  erb:"apartments/new"
+end
+
+get '/apartments/1/tenants' do
+  erb:"apartments/1/tenants"
```
andrewsunglaekim, line 25: read above for this route as well.

PR #13, opened 2015-11-04
Rachel solution - All parts

ddayporter:
Comfort: 5
Completeness: 4

Original PR 11/4: I feel pretty good about this, but would love any feedback if you see where the code could be cleaner or more efficient. UPDATE 11/5: I still feel pretty comfortable but cannot say completeness is 5. I took your advice and worked on it for a couple hours but don't have time to finish it. Since I finished the lab today (up to but not including the bonus) I think I do get the concepts pretty well, I just had a hard time working quickly enough to get the Landlord app up to the same level. One thing I couldn't figure out, my "create new apartment" form didn't work.. I got the "new artist" form to work on Tunr but for some reason it's not adding my new apartment to the list. Probably some small thing missing, ran out of time to troubleshoot..


app.rb
```diff
@@ -0,0 +1,49 @@
...
+require_relative "models/apartment"
+require_relative "models/tenant"
+
+get '/' do
+ erb :index
+end
+
+get '/apartments' do
+  @apartments = Apartment.all
+  @tenants = Tenant.all
```
mattscilipoti, line 18: Why are we retrieving all the Tenants from the db?

app.rb
```diff
@@ -0,0 +1,49 @@
...
+end
+
+get '/apartments/new' do
+  erb :"apartments/new"
+end
+
+get '/apartments/:id' do
+  @apartments = Apartment.all
+  @tenants = Tenant.all
+  @id = params[:id].to_i
```
mattscilipoti, line 29: You have the right pieces, just not put together quite right.  This route is expecting to show a single apartment.  

```
@apartment = Apartment.find(params[:id]) # rails takes care of string/int conversion
```

Then, use @apartment in the show view.

app.rb
```diff
@@ -0,0 +1,49 @@
...
+  erb :"apartments/show"
+end
+
+post '/apartments' do
+  @id = params[:id]
```
mattscilipoti, line 34: This is the create action.  A new Apartment doesn't have an id, yet.  You just need to create the apartment and redirect to a page that will show the new apartment (usually show or index).  You chose index.

mattscilipoti:
:+1:

PR #15, opened 2015-11-04
Tsolow solution

solowt:
comfort 4
completeness 5

Haven't started bonus yet

edit- added bonus app inside the app.rb file.  plan to add comments tomorrow morning and maybe make revisions.

last edit-added valid string checking to make sure apts/tenants actually exist before trying to modify them.  also added comments.


nolds9:
:+1: Awesome job tackling the bonus!

PR #16, opened 2015-11-04
landlords so far

Ly900:
Completeness: 3?
Comfort: 1

I am lacking the understanding of the big picture which is causing me confusion when working on this assignment. I also cannot remember all the steps we did in class because it was information overload today. Hopefully, things will sink in as we continue.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,4 @@
+source 'https://rubygems.org/'
+
+gem 'sinatra'
+gem 'sinatra-contrib'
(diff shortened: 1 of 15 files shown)


andrewsunglaekim:
I'm sorry you feel that way Ly! Your code, however, is indicative of more a 4.5 -5 in completeness. I imagine the comfort level you're feeling for this assignment largely stems from the amount of volume you've learned thus far in the last week. I was talking with your earlier and you expressed to me that you did understand everything a bit better as well. Anyways great work on this assignment, let me know if you  have any questions.

PR #19, opened 2015-11-04
Landlord Homework

lewis2ba:
Comfort - 4/5
Completeness 5/5

This homework went well! Although I'm not sure I understand the difference between "Updates attributes using attribute helper methods for one of the objects you've created" and  "Updates an object using the update methods". Thanks!


RobertAKARobin:
That's the difference between these two snippets:

``` rb
@apartment.update(address: "01 North Pole")
```

``` rb
@apartment.address = "01 North Pole";
@apartment.save
```

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,44 @@
+get "/apartments" do
+  apartments = Apartment.all
+  @apartments = apartments
```
RobertAKARobin, line 3: Why not just do `@apartments = Apartment.all`?

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+end
+
+post "/apartments/new" do
+  apartments = Apartment.all
+  @address = params[:address]
+  @monthly_rent = params[:monthly_rent]
+  @sqft = params[:sqft]
+  @num_beds = params[:num_beds]
+  @num_baths = params[:num_baths]
+  new_apartment = Apartment.create(address: @address, monthly_rent: @monthly_rent, sqft: @sqft, num_beds: @num_beds, num_baths: @num_baths)
```
RobertAKARobin, line 20: None of these really need to be instance variables since you're not referencing them in a view. In fact, they don't really even need to be variables! I'd just do `Apartment.create(address: params[:address]...)`. It'd be a lot less to type!

RobertAKARobin:
Looks good!

PR #21, opened 2015-11-04
Samir HW - Landlord 

samir-acle:


models/apartment.rb
```diff
@@ -0,0 +1,28 @@
...
+end
+
+# aprt = Apartment.new
+# aprt.add_tenant 'bob',23,'male'
+
+# puts aprt
+
+binding.pry
+
+puts "dlfhifsdf"
```
RobertAKARobin, line 28: Ah, yes. dlfhifsdf.

PR #23, opened 2015-11-04
landlord hw mati

matiamin:
completeness: 5
comfort: 4


app.rb
```diff
@@ -0,0 +1,14 @@
...
+
+require_relative "db/connection" # require the db connection file that connects us to PSQL, prior to loading models
+require_relative "models/apartment" # require the Student class definition that we defined in the models/student.rb file
+require_relative "models/tenant" # require the Student class definition that we defined in the models/student.rb file
+
+
+# This will put us into a state of the pry REPL, in which we've established a connection
+# with the wdi database and have defined the Student Class as an
+# ActiveRecord::Base class.
+binding.prys
```
adambray, line 14: Does this work with a plural `prys` at the end? Even if it does, you should use the singular `binding.pry`

db/models/apartment.rb
```diff
@@ -0,0 +1,3 @@
+class Apartment
```
adambray, line 1: Did you commit the right branch? This file looks incomplete.

adambray, line 1: Oh, and we usually keep our `models` folder separate (outside of) our `db` folder.

db/query_exercises.sql
```diff
@@ -1,12 +1,24 @@
...
 -- Find just the ids for all apartments
+SELECT apartment_id FROM tenants;
 -- Find all info for apartments whose price is greater than $2300
+SELECT * FROM apartments WHERE monthly_rent > 2300;
 -- Geriatric Birthday! Update all tenants whose age is 90 to be 91
+UPDATE tenants SET age = 91 WHERE age = 90;
 -- Change all tenants ages to increase by 1
+UPDATE tenants SET age = age+1;
 -- Find all tenants who live in an apartment that costs more than $2300
+SELECT * FROM tenants WHERE monthly_rent > 2300;
```
adambray, line 24: Did you try this command? I it won't work, since tenants don't have a monthly rent column, that's on apartments... how can we query both apartments and tenants at the same time?

db/models/apartment.rb
adambray, line 1: Oh, looking below I see this is a 'dead' file since you have another model elsewhere. You should be careful and not leave old dead code lying around! It can get very confusing as our projects get bigger!

models/apartment.rb
```diff
@@ -0,0 +1,24 @@
...
+#     @monthly_rent = monthly_rent
+#     @sqft = sqft
+#     @num_beds = num_beds
+#     @num_baths = num_baths
+#     @renters = renters
+#
+#   end
+#   def add_tenant
+#     while renters <= num_beds
+#       renters += 1
```
adambray, line 21: This isn't wrong per se, but the intention was for renters to be an array of Tenant objects, not just a number counting them.

adambray, line 21: (i.e. to model how AR does relationships)

models/apartment.rb
```diff
@@ -0,0 +1,24 @@
+class Apartment < ActiveRecord::Base
+has_many :tenants
+end
```
adambray, line 3: :+1:

models/tenant.rb
```diff
@@ -0,0 +1,12 @@
+class Tenant < ActiveRecord::Base
+belongs_to :apartment
+end
```
adambray, line 3: :+1:

adambray:
Looks good!

PR #24, opened 2015-11-04
Cpgruber solution (wednesday)

cpgruber:
Comfort_level: 4
Completeness: 4

Felt better setting up classes and associations in AR than in plain Ruby. Still unsure how AR and Ruby might mix, however. For instance, if you're using AR can you define class methods in Ruby too?


models/apartments.rb
```diff
@@ -0,0 +1,18 @@
...
+  #   @sqft = sqft
+  #   @num_beds = num_beds
+  #   @num_baths = num_baths
+  #   @renters = renters
+  # end
+  # def add_tenant (tenant)
+  #   if @renters.length < @num_beds
+  #     @renters << tenant
+  #   end
+  # end
```
jshawl, line 17: if you uncomment this method `add_tenant`, you will have defined an instance method on your active record inherited Apartment class. Then you'll be able to do things like:

``` rb
apt = Apartment.first
tenant = Tenant.first
apt.add_tenant(tenant)
```

Since this is an association, I would rely on active record methods to accomplish this.

A more practical application would be something like:

``` rb
class Person < ActiveRecord::Base
  # in the db, person has a first name and last name columns
  def full_name
    self.first_name + " " + self.last_name
  end
end
```

jshawl:
:+1: Code looks great! Let me know if you'd like to chat more about model methods

models/apartments.rb
cpgruber, line 17: That's what I wasn't sure about-- if I could define class methods in addition to AR methods. Based on your comment, it sounds like you can. So you're saying that an `add_tenant` method should be accomplished by updating the apartment attribute for a tenant in AR, rather than defining the method in the ruby class, whereas something as trivial as concatenating first and last name should be done in the ruby class out of convenience... right?

jshawl, line 17: apologies for the delay - you're absolutely right! keep up the great work

PR #27, opened 2015-11-04
Alexis_Landlord_part1

ajohnson052:
Completeness: 5
Comfort: 2.5

I'm not comfortable with the organization of files for the Sinatra segment (e.g. when pulling tenants of a specific apartment), but I'm hoping this will make more sense after working with Active Record.

### DIFF
--- a/controllers/apartments.rb
+++ b/controllers/apartments.rb
@@ -1,0 +1,45 @@
+require "pry"
+
+get "/apartments" do
+  @apartments = Apartment.all
+  erb :apartment_index
+end
+
+get "/apartments/new" do
+  erb :apartment_new
+end
+
+get "/apartments/search" do
+  apartment_id = Apartment.find_by(address: params[:address]).id
+  redirect "/apartments/#{apartment_id}"
+end
+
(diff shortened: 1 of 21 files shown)


ajohnson052:
Comfort: 4
Completeness: 4

I don't think that I did the post requests in the way that the instructions intended, but I think they accomplish the same task.

controllers/apartments.rb
```diff
@@ -0,0 +1,45 @@
...
+end
+
+get "/apartments/:id" do
+  @apartment = Apartment.find(params[:id])
+  erb :apartment_detail
+end
+
+post "/apartments" do
+  new_apt = Apartment.create(address: params[:address], monthly_rent: params[:monthly_rent], sqft: params[:sqft], num_beds: params[:num_beds], num_baths: params[:num_baths])
+  redirect "/apartments/#{new_apt.id}"
```
jsm13, line 33: Not only does it work, I think it's better. Nice!

jsm13:
:+1: Awesome! Looks really great

PR #29, opened 2015-11-04
initial commit. apartment/ tenant rb files, sql

tkentor:
comfort: 2.5
completeness: 4?

Comfort low because I got a lot of help on it, but it makes a lot more sense than before I started the exercise. I think I did everything but not entirely sure, that's why completeness = 4? .


tkentor:
comfort: 1.5
completeness: 2.5

jsm13:
:+1:  Good start here. How did review this weekend go?

PR #30, opened 2015-11-04
Mia Active Record Update

mcgrodsky:
Comfort 3/5 - I really like using Active Record and feel good about it, but had some issues (see below) 

Complete 3/5 - Active Record is set up and working, however i did not complete all of the assigned tasks because I was having some issues when opening ruby app.rb. I felt like the commands i was using were accurate, but i wasn't able to do things like assign a variable name to a new tenant and store it.


app.rb
```diff
@@ -0,0 +1,14 @@
+require "bundler/setup" # require all the gems we'll be using for this app from the Gemfile. Obviates the need for `bundle exec`
+require "pg" # postrgres db library
+require "active_record" # the ORM
+require "pry" # for debugging
+
+require_relative "db/connection" # require the db connection file that connects us to PSQL, prior to loading models
+require_relative "models/tenant" # require the Student class definition that we defined in the models/student.rb file
+require_relative "models/apartment"
+require_relative "db/seeds.rb"
```
jshawl, line 9: This line will execute your seeds file

db/seeds.rb
```diff
@@ -0,0 +1,41 @@
...
+dupont_apt.tenants.create(name: "Sally Fessler", age: 45, gender: "Female")
+dupont_apt.tenants.create(name: "Sara Robatille", age: 35, gender: "Female")
+
+# sally = Tenant.create(name: "Sally Mildren", age: 50, gender: "Female")
+
+# # query for all objects of the Tenant class, store it in a variable
+tenant = Tenant.all
+#
+# # query for all instances of the Tenant class that belong to the first Apartment you created
+colheights =
```
jshawl, line 33: maybe this is causing a syntax error that's preventing `ruby app.rb` from running?

app.rb
jshawl, line 9: on second thought, I would remove line 9, and explicitly run the seeds file when you need to seed with `ruby seeds.rb`

jshawl:
:+1: Let me know if you'd like to work through this!

PR #32, opened 2015-11-04
Brendan mahoney solution

BMahoney08:
Comfort - 4/5
Completeness - 3/5

I feel fairly comfortable with the code, it just takes me a long time to write, which is why I was unable to complete the HW. My only question would be regarding forms used to create/update a record. For my "update" functionality on tunr I passed strings that corresponded with with keys into params.

```
put "/artists/:id" do
  @artist = Artist.find(params[:id])
  @artist.update(name: params["name"], nationality: params["nationality"])
  redirect "/artists"
end

```

However, for my create functionality on landlord I passed the keys as keys.

```
post "/apartments" do
  @all_apartments = Apartment.all
  @new_apartment = Apartment.create(address: params[:address], monthly_rent: params[:monthly_rent],         sqft: params[:sqft], num_beds: params[:num_beds], num_baths: params[:num_baths])
  @all_apartments << @new_apartment
  redirect "/apartments"
end

```

Is this just a matter of choice? Or is there a best practice?


RobertAKARobin:
I'd stick with symbols, unless you _have_ to use strings for some reason. This is convention, probably mostly because symbols are much faster than strings. The reasons these both worked is that Ruby will sometimes convert strings to symbols for you, and vice-versa, when you're using them as keys. I say "sometimes" because I'm not entirely sure when!

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,53 @@
+require "pry"
```
RobertAKARobin, line 1: If you're `require_relative`-ing multiple files together, you actually only need to `require` Pry in one of them. If File A requires File B, and File B requires Gem C, then File A also ends up with Gem C.

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+get "/apartments" do
+@apartments = Apartment.all
+# binding.pry
+erb :"/apartments/index"
+end
+
+post "/apartments" do
+  @all_apartments = Apartment.all
+  @new_apartment = Apartment.create(address: params[:address], monthly_rent: params[:monthly_rent], sqft: params[:sqft], num_beds: params[:num_beds], num_baths: params[:num_baths])
+  @all_apartments << @new_apartment
```
RobertAKARobin, line 12: I personally prefer `.push`, but same thing.

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+
+get "/apartments/edit/:id" do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/edit"
+end
+
+put "/apartments/:id" do
+  @apartment = Apartment.find(params[:id])
+  @apartment.update(address: params[:address], monthly_rent: params[:monthly_rent], sqft: params[:sqft], num_beds: params[:num_beds], num_baths: params[:num_baths])
+  redirect "/apartments/#{@apartment.id}"
```
RobertAKARobin, line 47: It looks like you've gotten very good at seeing the patterns and commonalities between all these routes. This is great!

RobertAKARobin:
Really nicely done, Brendan!

PR #33, opened 2015-11-04
Britton Landlord Solution

brittonwalker:
Comfort: 5
Completeness: 5

I am still going to do the bonus.


nolds9:
:+1: Great job!

PR #37, opened 2015-11-04
Tim Gregg landlord HW

tomiteo:
3.9/5

Was feeling really good about this whole assignment until I tried passing in 

`Apartment.new(address: "19 rue Italia", monthly_rent: 500, sqft: 400, num_beds: 3, num_baths: 2)`

and got returned:

`ArgumentError: wrong number of arguments (1 for 5)`

I can't find the bug, but I've written the skeleton of all the answers, and I'll ask for help at office hours during lunchtime.

### DIFF
--- a/models/apartment.rb
+++ b/models/apartment.rb
@@ -6,7 +6,7 @@ class Apartment < ActiveRecord::Base
   attr_accessor :num_baths
   attr_accessor :renters
 
-  def initialize (address, monthy_rent, sqft, num_beds, num_baths, renters)
+  def initialize (address, monthy_rent, sqft, num_beds, num_baths)
     @address=address
     @monthy_rent=monthy_rent
     @sqft=sqft
@@ -21,5 +21,5 @@ class Apartment < ActiveRecord::Base
     end
   end
 
-  has_many :songs
+  has_many :tenants
 end
--- a/models/apartment.rb
+++ b/models/apartment.rb
@@ -1,6 +1,4 @@
-require "pry"
-
-class Apartments
+class Apartments < ActiveRecord::Base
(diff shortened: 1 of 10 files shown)


models/apartment.rb
```diff
@@ -0,0 +1,25 @@
...
+  attr_accessor :renters
+
+  def initialize (address, monthy_rent, sqft, num_beds, num_baths)
+    @address=address
+    @monthy_rent=monthy_rent
+    @sqft=sqft
+    @num_beds=num_beds
+    @num_baths=num_baths
+    @renters=Array.new(renters)
+  end
```
jshawl, line 16: This is what's causing the error you mentioned in the comment of your pull request. Remove the initialize method entirely and let active record handle object instantiation

models/apartment.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def add_tenant(tenant)
+    if @renters.length < @num_beds
+      @renters.push(tenant)
+    end
+  end
+
+  has_many :tenants
```
jshawl, line 24: this works! but convention is to specify associations at the top of the file

models/tenant.rb
```diff
@@ -0,0 +1,13 @@
+class Tenant < ActiveRecord::Base
+  attr_accessor :name
+  attr_accessor :age
+  attr_accessor :gender
+
+  def initialize name, age, gender
+    @name=name
+    @age=age
+    @gender=gender
+  end
```
jshawl, line 10: remove initialize method here as well

jshawl:
Keep feelin good! your code looks great - just need to remove the initialize methods in your classes that inherit from active record

PR #39, opened 2015-11-04
Jmoon solution

mooniker:
Completeness 3/5
Comfort: 2/5

After looking at the solution, I think I misunderstood how I'm supposed to create the directory tree and pages under views.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -2,13 +2,6 @@ require 'sinatra'
 require 'sinatra/reloader'
 # require 'haml'
 
-get '/' do
-  # @name_string = "#{params[:last_name]}, #{params[:first_name]} #{params[:last_name]}!"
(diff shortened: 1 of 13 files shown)


app.rb
```diff
@@ -0,0 +1,67 @@
...
+    attr_accessor :address
+
+    def initialize address
+      @address = address
+    end
+
+  end
+  @apartments = [
+    Apartment.new("123 1st Ave"), Apartment.new("234 2nd St")
+  ]
```
andrewsunglaekim, line 31: Here, we'll just do an AR query instead

andrewsunglaekim:
Really liked that you used your apartment class and created 2 apartments to generate. Your code is indicative that you have greater than what you recorded for comfort and completeness. In fact completeness appears to be a 5.

PR #41, opened 2015-11-04
Matthew murgia solution

murgia:
comfort:3
complete:4

The first two worked well for me but I had trouble with Sinatra.


andrewsunglaekim:
:+1:  Looks pretty good for the most part. And for now, since we didn't have AR yet, all we expected was some dummy content in the views.

PR #42, opened 2015-11-04
initial commit

jacohen10:
comfort: 3.5
completeness: 5 (skipped the manual entry one, but Erica said to do that)

I think I understand everything in theory, which is a good start, but I think I simply need more time to practice everything. Also, I think classroom 1's schedule is a little out of order, which isn't a huge deal, but I think it made it a little harder to put everything together. I'm hoping to make a lot of progress tomorrow at lab.


amaseda:
Yeah, Classroom 1 is doing this assignment out of order with Sinatra before ActiveRecord. That being said, you don't need ActiveRecord to do the Sinatra portion.

hw/app.rb
```diff
@@ -0,0 +1,55 @@
...
+# - The homepage should list several menu options:
+#   * List all apartments (a link to `GET /apartments`)
+
+
+get '/apartments' do
+  erb :apartment_index
+end
+
+#   * View an apartment's details(a link to `GET /apartments/1`)
+get 'apartments/:id' do
```
amaseda, line 18: You'll want to move your `apartments/:id` route below `apartments/new`. Otherwise, your controller is always going to interpret "new" as an apartment id, which will always return an error because "new" is a string.

hw/app.rb
```diff
@@ -0,0 +1,55 @@
...
+# - The route `GET /apartments/1/tenants` should list all tenants for 1 apartment.
+get 'apartments/:id' do
+  erb :apartment_details
+end
+
+# - The route `GET /apartments/1/tenants/new` should show a form for adding a new tenant.
+#   * Make sure to get the appropriate input from the user to create your person as per schema
+get '/apartments/:id/tenants/new'
+  erb :new_tenant
+end
```
amaseda, line 55: Guessing you might have just left these in here, but you don't need these duplicate routes.

amaseda:
Great job so far. And yes, I think a lot of the things you learned this week will solidify during the lab. Will take another look at the remainder of Landlord on Friday. As always, let me know if you have any questions in the meantime.

:+1:

PR #43, opened 2015-11-04
First upload with SQL seed injected

Cajam:


models/apartment.rb
```diff
@@ -0,0 +1,21 @@
+class Apartment
```
jshawl, line 1: Make sure to inherit from `ActiveRecord::Base` to interact with the db

models/apartment.rb
```diff
@@ -0,0 +1,21 @@
...
+  attr_accessor :renters
+  attr_reader :address, :monthly_rent, :sqft, :numb_beds, :num_baths
+
+  def initialize(address)
```
jshawl, line 5: once you inherit from AR::Base, you can remove the initialize method and let Active Record handle object instantiation

jshawl:
:+1: You're off to a great start! Let me know if you'd like to work through the Active Record Integration together

PR #44, opened 2015-11-04
alex's_solution

ascreven:
comfort: 4
completeness: 4

I was confused how to make a function in the apartment.rb file and still incorporate ActiveRecord


ascreven:
Completeness: 3
Comfort: 4

db/query_exercises.sql
```diff
@@ -1,12 +1,35 @@
...
--- Retrieve all info on all tenants in apartment with id 20
--- Retrieve all info on all tenants in apartment with ids 20 or 15
--- Delete all tenants whose age is greater than 65
--- Create one new tenant, put them in any apartment you want
--- Find just the ids for all apartments
--- Find all info for apartments whose price is greater than $2300
--- Geriatric Birthday! Update all tenants whose age is 90 to be 91
--- Change all tenants ages to increase by 1
--- Find all tenants who live in an apartment that costs more than $2300
+Retrieve all info on all tenants
```
mattscilipoti, line 1: The `--` at the from t of these lines marks them as a comment. If you leave each "exericse request" line as a comment, then you can run this file (thru psql) and see the results.

db/seeds.rb
```diff
@@ -0,0 +1,67 @@
+require 'bundler/setup'
+require "pg"
+require "active_record"
+require "pry"
+
+require_relative "../models/tenant"
+require_relative "../models/apartment"
+
+require_relative "../connection"
```
mattscilipoti, line 9: I recommend connecting to the database prior to requiring the models, because they depend on the database.

db/seeds.rb
```diff
@@ -0,0 +1,67 @@
...
+kamille.tenants.create(name: "Valentin Keebler", age: 74, gender: 'Male')
+damien.tenants.create(name: "Ms. Piggy", age: 24, gender: "Female")
+koch.tenants.create(name: "Eryn Lynch", age: 12, gender: 'Female')
+hal.tenants.create(name: "Tiara Conn", age: 3, gender: 'Female')
+buford.tenants.create(name: "America Pollich", age: 82, gender: 'Female')
+tyree.tenants.create(name: "Mylene Krajcik", age: 25, gender: 'Female')
+runte.tenants.create(name: "Josh Gottlieb", age: 8, gender: 'Male')
+marvin.tenants.create(name: "Sherwood Stiedemann", age: 72, gender: 'Female')
+leta.tenants.create(name: "Katarina Kunde", age: 6, gender: 'Female')
+
```
mattscilipoti, line 60: That's a lot of seed data.  Looks like you had fun.  I thought this might reference something fun, but I couldn't find a connection.

mattscilipoti:
:+1: 

The Apartment class is just like any other ruby class.  Just add `def method_name`.  Make sure you are NOT overwriting a method that is added by `has_many` (unless you want to).

views/apartment.erb
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
mattscilipoti, line 1: This belongs in app/views/layout.erb.  The views for each model are just html within the `<body>`.  Remember, these expand within the location where the `<%= yield %>` command is.

mattscilipoti:
I'm curious, what method did you want to add to Apartment?

PR #46, opened 2015-11-04
M. Shofner - Week 4 HW - Landlord (Wednesday. AR, no Sinatra)

MarkhamShofner:
Comfort - 3
Completeness - 4

Probably not the cleanest.

### DIFF
--- a/db/query_exercises.sql
+++ b/db/query_exercises.sql
@@ -1,12 +1,35 @@
 -- Retrieve all info on all tenants
+SELECT * FROM tenants;
+
 -- Retrieve the name, age, and gender of all tenants
+SELECT name, age, gender FROM tenants;
+
 -- Retrieve all info on all tenants older than 65
+SELECT * FROM tenants WHERE age > 65;
+
 -- Retrieve all info on all tenants in apartment with id 20
+SELECT * FROM tenants WHERE apartment_id = 20;
+
 -- Retrieve all info on all tenants in apartment with ids 20 or 15
+SELECT * FROM tenants WHERE apartment_id = 20 or apartment_id = 15;
+
 -- Delete all tenants whose age is greater than 65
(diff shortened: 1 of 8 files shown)


db/query_exercises.sql
```diff
@@ -1,12 +1,35 @@
...
+SELECT * FROM apartments WHERE monthly_rent > 2300;
+
 -- Geriatric Birthday! Update all tenants whose age is 90 to be 91
+UPDATE tenants SET age = 91 WHERE age = 90;
+
 -- Change all tenants ages to increase by 1
+UPDATE tenants SET age = age + 1;
+
 -- Find all tenants who live in an apartment that costs more than $2300
+SELECT * FROM tenants JOIN apartments ON tenants.apartment_id=apartments.id WHERE monthly_rent > 2300;
```
mattscilipoti, line 35: FYI: this will return all the columns from both the tenants and apartments tables.

seeds.rb
```diff
@@ -0,0 +1,55 @@
...
+ap2_tenants = ap2.tenants
+
+# ○ Updates attributes using attribute helper methods for one of the objects you've created
+ap3.monthly_rent = 3420
+
+# ○ Saves an object that you updated using attribute helpers to the Database
+ap3.save
+
+# ○ Updates an object using the update methods
+ten5 = Tenant.all[4]
```
mattscilipoti, line 43: I recommend retrieving the specific record using ActiveRecord: e.g. `Tenant.find(id)`.  For two reasons:
- `Tenant.all[4]` returns an array of all the tenants form the db and then selects one.  What if we had 200,000 tenants?
- That code selects the record in the 5th position.  The order of records is non-deterministic (unless you define the order).  Even then, it's tough to get it right every time.

mattscilipoti:
:+1:

PR #47, opened 2015-11-04
Initial Commit

toryburgett:
Incomplete, still working on this assignment

### DIFF
--- a/db/schema.sql
+++ b/db/schema.sql
@@ -1,0 +1,31 @@
+-- Tenants table (with the following attributes):
+-- id
+-- name
+-- age
+-- gender
+-- apartment_id
+--
+-- Apartments table (with the following attributes):
+-- id
+-- address
+-- monthly_rent
+-- sqft
+-- num_beds
+-- num_baths
+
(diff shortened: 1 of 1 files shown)


db/schema.sql
```diff
@@ -0,0 +1,31 @@
...
+-- sqft
+-- num_beds
+-- num_baths
+
+CREATE TABLE tenants (
+  tenants_id_seq SERIAL NOT NULL PRIMARY KEY,
+  name VARCHAR NOT NULL,
+  age INT,
+  gender VARCHAR,
+  apartment_id INT NOT NULL
```
amaseda, line 21: Keep in mind that if you set `apartment_id` to `NOT NULL` then a tenant must always be attached to an apartment. Nothing wrong with that, but know that you can't have any "floating" tenants now.

amaseda:
Good job with the schema so far. Let me know if you have any questions about the topics that have been covered since that lesson (Ruby Classes, ActiveRecord, Sinatra REST). Will take a look at the rest of the assignment on Friday!

:+1:

PR #48, opened 2015-11-04
Dt solution AR Landlord HW

dttruong:
comfort: 3.5
completeness: 5

it took me a while to connect all of the .rb files to the app.rb, etc etc.  I seem to understand it though for most parts

### DIFF
--- a/tenant.rb
+++ b/tenant.rb
@@ -1,0 +1,8 @@
+require "pry"
+
+class Tenant
+
+
+
+
+end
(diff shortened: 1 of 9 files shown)


tenant.rb
```diff
@@ -0,0 +1,8 @@
+require "pry"
```
nolds9, line 1: Pretty sure we don't need this file...

nolds9:
:+1:  Great job!

PR #52, opened 2015-11-04
Chrystal's Homework

CBanks813:
Still in progress.

### DIFF
--- a/Apartment.rb
+++ b/Apartment.rb
@@ -1,0 +1,3 @@
+class Apartment < ActiveRecord::Base
+  has_many :tenant
+end
--- a/Tenant.rb
+++ b/Tenant.rb
@@ -1,0 +1,3 @@
+class Tenant < ActiveRecord::Base
+  belongs_to :apartment
+end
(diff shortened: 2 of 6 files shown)


Apartment.rb
```diff
@@ -0,0 +1,3 @@
+class Apartment < ActiveRecord::Base
```
adambray, line 1: The name of the class here is correct, in that it's Capitalized, but note that the name of the file should should `apartment.rb`. (note the lowercase). Our file names will almost always be lower_snake_case.

Apartment.rb
```diff
@@ -0,0 +1,3 @@
...
+  has_many :tenant
```
adambray, line 2: This should be `has_many :tenants`. A tip is that these methods almost always make sense in terms of english, has_many implies plural, so `:tenants`.

Tenant.rb
```diff
@@ -0,0 +1,3 @@
+class Tenant < ActiveRecord::Base
```
adambray, line 1: This all looks good, except again, note that the file name should be lower case.

adambray:
Looks pretty good.

PR #54, opened 2015-11-05
Wednesday Pull Request

KoryJCampbell:


models/apartment.rb
```diff
@@ -0,0 +1,13 @@
+class Apartment < ActiveRecord::Base
+end
+
+def initialize
```
amaseda, line 4: You actually don't need an initialize method when inheriting from ActiveRecord.

amaseda:
Good start. Will take another look when you submit the Landlord assignment in full Friday.

And a reminder to always include score for Comfort and Completeness, as well as at least one question. You'll find the latter to be more helpful than you think! Also indicates what I should comment on.

:+1:

PR #59, opened 2015-11-06
Rebecca solution

rebeccaestes:
completeness: 3.5
comfort: 4

I made the page structure a lot different than the readme instructions instructed, but I didn't realize that til partway through. They still make intuitive sense. I also couldn't get the apartments to be editable or deletable - you can view the forms OK, but you get a Sinatra error page when you submit, and the apartments are unchanged.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -3,8 +3,45 @@ require "pg"
 require "active_record"
 require "pry" 
 
-require_relative "connection" 
-require_relative "../models/apartment" 
-require_relative "../models/tenant"
+require_relative "db/connection" 
+require_relative "models/apartment" 
+require_relative "models/tenant"
 
-binding.pry
+
+puts "Do you want to add a tenant, or an apartment?"
+
+def add_tenant
+	puts "Name?"
+	tenant_name = gets.chomp
+	puts "Age?"
+	tenant_age = gets.chomp.to_i
+	puts "Gender?"
+	tenant_gender = gets.chomp
+	puts "Apartment ID?"
+	tenant_apt = gets.chomp.to_i
+	puts "OK. I'll add #{tenant_name}, a #{tenant_gender} age #{tenant_age}, to apartment #{tenant_apt}."
+	# if ()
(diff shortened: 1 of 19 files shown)


app.rb
```diff
@@ -0,0 +1,94 @@
...
+put "/apt_edit/:id" do
+    @id = params[:id]	
+    @apt = Apartment.find(@id)
+	@address = params[:address]
+	@monthly_rent = params[:monthly_rent].to_i
+	@sqft = params[:sqft].to_i
+	@num_beds = params[:num_beds].to_i
+	@num_baths = params[:num_baths].to_i
+	@apt.update(address: @address, monthly_rent: @monthly_rent, sqft: @sqft, num_beds: @num_beds, num_baths: @num_baths)
+	redirect "/apartment/#{@id}"
```
ebirving, line 86: Also, `apartments/`, not `apartment/`, is the ditty that Sinatra knows.

app.rb
```diff
@@ -0,0 +1,94 @@
...
+	@id = params[:id]
+	erb :tenant_new
+end
+
+get "/tenant_details/:id" do
+	@apt = Apartment.find(params[:id])
+	erb :tenant_details
+end
+
+get "/apt_edit/:id" do
```
ebirving, line 72: The root of your problem with update and delete is here in edit. You reference `@id` in the action of your form, but you don't define it in the corresponding controller ( `get "/apt_edit/:id`...). When submitting either of the forms in that view triggers the assembly of the route from the method and action specified in your `form` tags, it doesn't know what to do with `@id`, so it skips it, and tries to `put` or `delete`to `/apt_edit/`. If you use `@apt.id` in place of `@id` in the forms, however...

ebirving:
Good work, Rebecca. Take a look at my comments on edit/delete and let me know if anything is unclear. I encourage you to take your time reading errors. There's a lot of information in there, particularly in Rails/Sinatra, if you pay attention to the details.

PR #60, opened 2015-11-06
ThomasS-solution

solowt:
comfort: 4
completeness: 5

I feel a lot more comfortable after going through this same process twice.  

The app.rb file still contains the cli app. The sinatra app is in the sinatra_app.rb file.   

I did notice that the information in the seeds.sql file seems a little off.  There's only 19 apartments but it assigns tenants to apartment 20, so if you make a new apartment, it gets a bunch of tenants right away.  I don't know if it was like that when I originally cloned it or if I deleted an insertion at some point.

Also, I didn't get around to adding comments or doing any formatting at all.  I might do some of that later tonight or tomorrow morning.


controllers/apartments_controller.rb
```diff
@@ -0,0 +1,40 @@
+# show the index page
+get "/apartments" do
+  @apt_list = Apartment.all
```
nolds9, line 3: While it won't change anything, convention is to use `@apartments`

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+end
+# show the new apt page
+get "/apartments/new" do
+  erb :"apartments/new"
+end
+# create a new apt
+put "/apartments" do
+  edit_params = params.except("_method") # remove unwanted _method key
+  edit_params.delete_if {|key, value| value == ""} # remove blank values from the hash
+  Apartment.create(edit_params)
```
nolds9, line 14: A little confused by the consistency of this controller action, If this is a create action, then you would use a `post '/apartments'` route, and then I would just change your naming to be consistent. Otherwise if this is an update action, you would find a specific apartment, save that to an instance, then call `.update` on that instance, passing in the `edit_params`

controllers/tenants_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  erb :"tenants/show"
+end
+# get the tenants in a specific apt
+get "/apartments/:id/tenants" do
+  @apt = Apartment.find(params[:id])
+  @tenant_list = Tenant.where(apartment_id: @apt.id)
+  erb :"tenants/show2"
+end
+# add a new tenant here
+put "/apartments/:id/tenants" do
```
nolds9, line 30: `post`

nolds9:
:+1: Great job on this! Please review inline comments for feedback, and let me know if you have any questions.

PR #62, opened 2015-11-06
Chris kibble sinatra views

ckib16:
Basic controller done, but ran out of time on functionality


controllers/apartment_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# SHOW
+get '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/show"
+end
+
+# EDIT
+get '/apartments/:id/edit' do
+  @artist = Apartment.find(params[:id])
```
RobertAKARobin, line 31: I think you mean `@apartment`! :)

controllers/apartment_controller.rb
```diff
@@ -0,0 +1,49 @@
...
+  erb :"apartments/edit"
+end
+
+# UPDATE
+put '/apartments/:id' do
+  @edited_apartment = params[:apartment]
```
RobertAKARobin, line 37: I'd call this `@apartment_params`... although, honestly, you don't really even need a variable for this, since you're not using it in your views.

PR #64, opened 2015-11-06
HW Daniel Alexander Sinatra DB

dan-ator:
comfort: 4.5
completeness: 5 (didn't get too far into the bonus)

### DIFF
--- a/controllers/apartments_controller.rb
+++ b/controllers/apartments_controller.rb
@@ -1,0 +1,46 @@
+get '/apartments' do
+  @apartments = Apartment.all
+  erb :"apartments/index"
+end
+
+get '/apartments/new' do
+  erb :"apartments/new"
+end
+
+post '/apartments/new' do
+  @address = params[:address]
+  @rent = params[:monthly_rent]
+  @sqft = params[:sqft]
+  @beds = params[:num_beds]
+  @baths = params[:num_baths]
+  Apartment.create(address:"#{@address}", monthly_rent:"#{@rent}", sqft:"#{@sqft}", num_beds:"#{@beds}", num_baths:"#{@baths}")
+  redirect "/apartments"
+end
+
+get '/apartments/:id' do
+  @apartments = Apartment.find(params[:id])
+  @tenants = Tenant.where(apartment_id: params[:id])
+  erb :"apartments/show"
+end
+
+get '/apartments/:id/edit' do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/edit"
+end
+
+put '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  @address = params[:address]
+  @rent = params[:monthly_rent]
+  @sqft = params[:sqft]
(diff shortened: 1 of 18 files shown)


controllers/apartments_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+get '/apartments/new' do
+  erb :"apartments/new"
+end
+
+post '/apartments/new' do
+  @address = params[:address]
+  @rent = params[:monthly_rent]
+  @sqft = params[:sqft]
+  @beds = params[:num_beds]
+  @baths = params[:num_baths]
```
adambray, line 15: No need to use instance vars here, a local var `baths` (etc) would do just fine. Instance vars are only needed if you're rendering a view, and need to pass that data to the view. A redirect doesn't render, it triggers a new request to a different route.

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+  @apartment.update(address: "#{@address}", monthly_rent:"#{@rent}", sqft:"#{@sqft}", num_beds: "#{@beds}", num_baths: "#{@baths}")
+  redirect "/apartments/#{@apartment.id}"
+end
+
+delete '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  @apartment.destroy
+  redirect"/apartments"
```
adambray, line 45: :+1:

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+end
+
+put '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  @address = params[:address]
+  @rent = params[:monthly_rent]
+  @sqft = params[:sqft]
+  @beds = params[:num_beds]
+  @baths = params[:num_baths]
+  @apartment.update(address: "#{@address}", monthly_rent:"#{@rent}", sqft:"#{@sqft}", num_beds: "#{@beds}", num_baths: "#{@baths}")
```
adambray, line 38: Why use string interpolation here? Just reference the vars directly (note you don't need instance vars, as I mentioned above:

``` ruby
apartment.update(address: address, monthly_rent: rent) # etc
```

views/apartments/edit.erb
```diff
@@ -0,0 +1,11 @@
+<h2>Edit the Apartment Information</h2>
+
+<form method="post" action="/apartments/<%= @apartment.id %>">
+  <input type="hidden" name="_method" value="put">
```
adambray, line 4: :+1:

views/apartments/index.erb
```diff
@@ -0,0 +1,7 @@
+<h1>List of Apartments</h1>
+
+<% @apartments.each do |apartment| %>
+  <li><a href="apartments/<%= apartment.id %>"<li><%= apartment.address %></a></li>
```
adambray, line 4: Nice!

adambray:
Looks great!

PR #65, opened 2015-11-06
Ly

Ly900:
Completeness: 2
Comfort: 3

Enjoyed working on what I could finish. Not familiar with forms yet. Was wondering if it's possible to sort those tenant names alphabetically by last name?


app.rb
```diff
@@ -0,0 +1,55 @@
...
+get "/" do
+  erb :home
+end
+
+# The homepage should list several menu options:
+# List all apartments (a link to GET /apartments)
+
+
+# View an apartment's details(a link to GET /apartments/1)
+get "/apartments/:id" do
```
andrewsunglaekim, line 23: seems you have some duplicate routes, these routes here and below are very similar to the ones defined in your controllers.

andrewsunglaekim:
All the views look great! You can absolutely sort, look at the Active record documentation for [ordering](http://guides.rubyonrails.org/active_record_querying.html#ordering). Seems like the forms you made would work, just need to make the request do something upon form submission.

PR #69, opened 2015-11-06
Hw

jnichols300:
completion 3/5? maybe 2?

Comfort 3-4/5

I didn't get a lot of the end part of the homework done.

I definitely understand the material, I'm just having trouble putting everything together and doing it on a timely basis. It takes me a lot of time to do some things. I think I'm going to start doing per your suggestion, looking at the solution earlier on for when I get stuck. That way I can probably get a firmer grasp of the concepts and hopefully make me a more efficient coder. I'll probably finish this soon and delete it and see if I can't do it over again with less help. That definitely seems like a much better way to learn!

### DIFF
--- a/controllers/tenants.rb
+++ b/controllers/tenants.rb
@@ -1,0 +1,17 @@
+get '/apartments/:id/tenants' do
+  @apartment = Apartment.find(params[:id])
+  erb :"tenants/index"
+end
+
+# new
+get '/apartments/:id/tenants/new' do
+  @apartment = Apartment.find(params[:id])
+  erb :"tenants/new"
+end
+
+# create
+post '/apartments/:id/tenants' do
+  apartment = Apartment.find(params[:id])
+  apartment.tenants.create!(params[:apartment])
+  redirect "/apartments/#{apartment.id}/tenants"
+end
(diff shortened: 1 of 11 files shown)


controllers/tenants.rb
```diff
@@ -0,0 +1,17 @@
+get '/apartments/:id/tenants' do
```
andrewsunglaekim, line 1: how are you observing your apartment views you created? seems like there's no controller for apartments yet.

db/app.rb
```diff
@@ -0,0 +1,25 @@
...
+require 'sinatra/reloader'
+
+require_relative 'db_connection'
+require_relative 'models/apartment'
+require_relative 'models/tenant'
+require_relative 'controllers/welcome'
+require_relative 'controllers/apartments'
+require_relative 'controllers/tenants'
+
+# get '/' do
```
andrewsunglaekim, line 11: the routes would work for those apartment views, you just need to query the database for objects.

andrewsunglaekim:
Thats a great idea. Just remember that the solution is there as guidance not as a crutch. There's a fine line between condensing 8 hours of understanding into 1 hour vs using code and not knowing what it does.  When working towards a solution for hw, make sure you try small incremental steps and get parts of the solution along the way. Feels like you wrote a bunch of code before the first bit of functionality can be assessed. If oyu have time, i encourage you to continue to work on this assignment. :+1:

PR #70, opened 2015-11-06
Samir HW - landlord

samir-acle:
complete: 4/5
comfort: 3.5/5
questions:  why does Apartment.all array have edited apartments move to the end of the array even though id has not changed?

### DIFF
--- a/controllers/apartments_controller.rb
+++ b/controllers/apartments_controller.rb
@@ -1,0 +1,27 @@
+get "/apartments" do
+  @apartments = Apartment.all
+  erb :"apartments/index"
+end
+
+get "/apartment/:id" do
+  @apartment = Apartment.find(params[:id])
+  @tenants = @apartment.tenants
+  erb :"apartments/show"
+end
+
+get "/apartments/new" do
+  erb :"apartments/new"
+end
+
+post "/apartments" do
+  @addess = params[:address]
+  Apartment.create(params.except("_method"))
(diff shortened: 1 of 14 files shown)


RobertAKARobin:
`Apartment.all` isn't exactly an array. It's an `ActiveRecord::CollectionProxy` or something, which inherits from Arrays, but is still different different. As such, it doesn't necessarily follow the same rules.

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+end
+
+get "/apartment/:id/edit" do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/edit"
+end
+
+post "/apartments" do
+  @addess = params[:address]
+  Apartment.create(params.except("_method"))
```
RobertAKARobin, line 23: Nice use of `except`!

views/apartments/edit.erb
```diff
@@ -0,0 +1,29 @@
+<form method="post" action="/apartment/<%= @apartment.id %>">
+  <input type="hidden" name="_method" value="put">
+  <fieldset>
```
RobertAKARobin, line 3: Challenge: could you make an ERB loop for this so you don't have to explicitly write out each fieldset/input?

RobertAKARobin:
Nicely done.

PR #71, opened 2015-11-06
HW submission

onesick:
I could not get the renters link showing renters information. I believe there is misconnection between tenants and find_by method using aprtment_id.
Also, I did not have time to verify database control by update, delete, create, and etc thru ActiveRecord. I will try these over the weekend.
Overall, REST, sinatra, and DB management is 4/5. Except, still I have to go thru my old works and copy and paste such as require 'something' and require-relative 'something'. As far as syntax goes, and why we need them, I believe I have some solid understandings. 
And for the last, can you please explain why when path is called the syntax is erb :"something"? If i put the syntax in erb: "something", colon position difference, the syntax is wrong. 

Thank you.


apartment.rb
```diff
@@ -0,0 +1,20 @@
+class Apartment <Tenant
```
jsm13, line 1: ~~We probably don't want apartment inheriting from tenant. We want our models to inherit from ActiveRecord::Base.~~

Found the model where you do have Apartment inherit from ActiveRecord::Base. Still don't want apartment to inherit from tenant here

apartment.rb
```diff
@@ -0,0 +1,20 @@
...
+  def initialize
+    @address;
+    @monthly_rent;
+    @sqft;
+    @num_beds;
+    @num_baths;
+    @renters;
+  end
+
+  def add_tenant
```
jmas13: We want to let ActiveRecord take care of managing relationships

jsm13:
:+1:  Nice job. How's transitioning to Rails going?

PR #72, opened 2015-11-06
Sinatra + AR homework

starrchen:
comfort: 4.5 -- I think I understand this but I am really bad at debugging Sinatra; I don't understand the error messages at ALL
completeness: 4.5

Wasn't able to finish debugging the edit/delete functions even when I went over the timebox (but I promise I know how to do those, they work on my tunr)

Question: Is there a way to get my app working with my show apartments working in the apartments controller file?

``` ruby
# view an apartment's details
get '/apartments/:id' do
  @apartment = Apartment.find(params[:id])
  # FOR ADDING TENANTS @tenants = Tenant.where(apartment_id: params[:id])
  erb :"apartments/show"
end
```

I've found that if I put this at the top, where I logically/chronologically write it, I get an error that says something like `No Apartment with id="new" found` when I begin to write the new apartment feature. is there a way for me to still have that at the top of my controller file or is this something I just have to be mindful of and leave at the bottom? Does leaving it at the bottom mean I could run into trouble in the future with the show function?

ALSO is it a general point that commented-out HTML in a .erb file will break everything? I tried to leave myself small comments and comment out old code but my app wouldn't work until I deleted those comments.

ALSO ALSO we didn't learn about using a controller folder, so I started with one controller file for Tunr and Landlord. I never opened a new one for Tunr and everything still worked, so what's the benefit of having multiple ones besides feature separation? Does it add time to the process? 

Will finish and add tenant functionality in the future.


amaseda:
Re: `No Apartment with id="new" found`

This happens when your `apartments/:id` route is placed above your `apartments/new` route.  Because of this, whenever you visit `apartments/new` in your browser, "new" is treated as a value for `id` and passed into `params[:id]`. You can fix this by moving `apartments/:id` below `apartments/new` in your controller.

amaseda:
Re: commented-out code in an `.erb` file

Nothing wrong with commenting out HTML in an `.erb` file, but if that commented out code includes ERB tags, you need to comment out those as well using the following syntax. Note the `#`...

```
<%#= apartment.address %>
```

starrchen:
@amaseda 
Re: No `Apartment with id="new" found` -- Yeah, I figured, but was wondering if there was a workaround for that. But I'll just try to remember to dot that!

Re: commented-out code in an `.erb` file -- I was playing around with the clownhat comments but I guess then I can't do actual HTML comments like `<-- comment -->`, even though that's how Atom default comments for `.erb` files. (Do you know if there's a way to change that?)

Also also (also), quick question about the instance variables in my controller. I end up having to repeat many of them for each method (e.g. `@address = params[:address]`, but is what would be the best way for me to not be so repetitive, if there is one?

amaseda:
Re: multiple controllers

Depending on what routes we want available on our site, we might create a controller for each model. So in the case of Tunr, if we wanted to have routes like `songs/new` or `songs:id` available to the user, we would put these in a separate `songs_controller.rb` file.

Additional controllers don't add time (if they do, it's negligible). All about separation of concerns.

amaseda:
Are you including a `!` in your HTML comments? `<!-- commented out code -->`

If it's still acting funky, I can take a look later.

starrchen:
@amaseda Yes, sorry, I mistyped in my comment, but I was just using Atom's comment shortcut to add in a few notes and they broke the code that way. Is it because it's technically still a Ruby file, so I need to do embedded Ruby comments instead of HTML ones?

amaseda:
You should be able to use plain ol' HTML comments to comment out HTML that does not include any ERB.

Otherwise, use this syntax for ERB: https://gist.github.com/somethingkindawierd/9120901

I'll be avail during office hours, so can take a look at the code in question then.

amaseda:
Re: debugging

Rails is a lot nicer when it comes to debugging. Usually it tells you exactly what file/line is messing things up.

amaseda:
Aaaaand re: your first question.

That show route looks right to me, including the commented-out code. Although I think a cleaner way to write that out would be...

```
get '/apartments/:id' do
  @apartment = Apartment.find(params[:id])
  @tenants = @apartment.tenants
  erb :"apartments/show"
end
```

amaseda:
Covered a lot here. Let me know if there's anything else. And come show me your code during office hours so we can sort out this ERB comment business. But great job with this assignment!

:+1:

starrchen:
@amaseda  - I thought there wasn't a `.tenants` attribute for the `Apartment` class but I'll double check that and look into it when I work on that functionality. Thanks!

amaseda:
@starrchen There is no `.tenants` attribute, but because we do this...

``` ruby
class Apartment < ActiveRecord::Base
  has_many :tenants     # emphasis on this association
end

class Tenant < ActiveRecord::Base
  belongs_to :apartment    # same here
end
```

We can now do...

``` ruby
# we can only call the below on individual instances of our models
@apartment.tenants
@tenant.apartment
```

PR #75, opened 2015-11-06
Landlord solution estennett

estennett:
comfort 3
completeness 3

I was pretty exhausted after yesterday's lab, but gave this homework a good shot last night with some other classmates.  I plan to complete it over the weekend as practice.


controllers/apartments_controller.rb
```diff
@@ -0,0 +1,36 @@
+get "/apartments" do
+  @all_apartments = Apartment.all
```
nolds9, line 2: While it won't change anything, convention is  to use `@apartments`

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  erb :"apartments/id"
+end
+
+put "/apartments/:id" do
+  @address = params[:address]
+  @rent = params[:monthly_rent]
+  @sqft = params[:sqft]
+  @beds = params[:num_beds]
+  @baths = params[:num_baths]
+  Apartment.create(address: @address, monthly_rent: @rent, sqft: @sqft, num_beds: @beds, num_baths: @baths)
```
nolds9, line 21: This looks like an update action, so I would recommend finding the specific apartment, then calling update on that instance and passing in the params. Or make this the create route and do: `post '/apartments'`

nolds9:
:+1: Great job! Please review inline comments for feedback. Let me know if you have any questions

PR #83, opened 2015-11-06
Tim Gregg Landlord Friday Sinatra/Rest

tomiteo:
4/5s

Feeling pretty good, primary struggles I had with this and tunr were the forms and deleting items.

I know it's not quite fully fledged out, but I did what I could.


app.rb
```diff
@@ -0,0 +1,11 @@
...
+require "pg" # postrgres db library
+require "active_record" # the ORM
+require "pry" # for debugging
+require "sinatra"
+require "sinatra/reloader"
+
+require_relative "db/connection"
+require_relative "models/apartment"
+require_relative "models/tenant"
+require_relative "controller.rb"
```
jshawl, line 11: the `.rb` is uneccesary here because you can only `require_relative` ruby files

controller.rb
```diff
@@ -0,0 +1,41 @@
+get '/' do
```
jshawl, line 1: consider moving these routes into separate controllers. What you have works! Just makes it a little easier to keep organized as your app grows in complexity.

controller.rb
```diff
@@ -0,0 +1,41 @@
...
+  erb :"tenants/index"
+end
+
+get '/apartments/new' do
+  @apartments=Apartment.all
+  erb :"apartments/new"
+end
+
+get '/apartments/:id' do
+  @id=params[:id].to_i
```
jshawl, line 21: ActiveRecord will handle the string to integer conversion so `.to_i` is not needed

controller.rb
```diff
@@ -0,0 +1,41 @@
...
+end
+
+get '/tenants/:id' do
+  @id=params[:id].to_i
+  @tenants=Tenant.all
+  erb :"tenants/id"
+end
+
+get '/apartments/:id/tenants' do
+  @id=params[:id].to_i-1
```
jshawl, line 33: what is the `-1` doing here?

views/apartments/id_tenants.erb
```diff
@@ -0,0 +1,6 @@
+<% Apartment.find(@id).tenants.each do |tenant| %>
```
jshawl, line 1: instead, create an instance variable in your controller. This will help separate your concerns

views/apartments/id_tenants_new.erb
```diff
@@ -0,0 +1,8 @@
+<form action="/new" method="post">
```
jshawl, line 1: action should be `/artists`. The creation is implied because you're using the POST http verb

jshawl:
:+1: You're off to a great start! With respect to deleting items, check out - https://github.com/ga-dc/tunr_sinatra/tree/lab-starter#part-32-artist-controller-and-views and scroll down to the "Updating and artist" bullet. Let me know if you'd like to work through this - I'm on office hours today!

PR #84, opened 2015-11-06
Jmoon solution

mooniker:
4/5 completeness
3/5 comfort

seems like a lot of routing and templating for such a simple app. after finally starting to get the hang of things, building out the functionality gets really repetitive. Maybe sinatra is more well suited to something more simple?


controllers/apartments.rb
```diff
@@ -0,0 +1,56 @@
...
+get '/apartments/:id/edit' do
+  @apt = Apartment.find_by(id: params[:id].to_s)
+  erb :'apartments/edit'
+end
+
+patch '/apartments/:id' do
+  id = params[:id].to_i
+  # binding.pry
+  Apartment.find_by(id: id).update(params.except('_method', 'splat', 'captures', 'id'))
+  redirect '/apartments'
```
andrewsunglaekim, line 34: whoa, what does except do? cool. splat, _method, captures?!

controllers/apartments.rb
```diff
@@ -0,0 +1,56 @@
+get '/apartments' do
```
andrewsunglaekim, line 1: :+1:

controllers/tenants.rb
```diff
@@ -0,0 +1,28 @@
+get '/tenants' do
```
andrewsunglaekim, line 1: :+1:

andrewsunglaekim:
Think you hit the nail right on the head with that comment! Rails will abstract alot of that repetition out. Sinatra is great for something simple though. It's more pure in a sense that, everything you want in a sinatra application is there because you coded it to be there.

PR #85, opened 2015-11-06
Landlord

jordanbrauner:
Comfort: 3
Completeness: 2

Comments: Have not started AR or Sinatra REST. I will finish those parts over the weekend but I do not require feedback on them.

### DIFF
--- a/apartment.rb
+++ b/apartment.rb
@@ -1,0 +1,32 @@
+require 'pry'
+
+class Apartment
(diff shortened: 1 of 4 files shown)


amaseda:
Looking good so far. Still happy to provide feedback on AR and Sinatra REST portions even if it's next week, since those portions cover the newer material.

:+1:

PR #86, opened 2015-11-06
Pz solution

pzhang87:
completion: 3
comfort: 3.5

getting there!  timeboxed 2.5 hours for this, I think I can actually finish up in a few hours over the weekend.  I think it's just a matter of pushing through at this point

### DIFF
--- a/controllers/apartments_controller.rb
+++ b/controllers/apartments_controller.rb
@@ -1,0 +1,19 @@
+require 'pry'
+
+get '/apartments' do
+  @apartments_all = Apartment.all
+  binding.pry
+  erb :"apartments/index"
+end
+
+get '/apartments/:apartment_id' do
+  "ur mom in this apartment"
+end
+
+get '/apartments/new' do
+  "ur mom in a new place"
+end
+
+get '/apartments/:apartment_id/tenants' do
+  "ur tenants do ur mom"
+end
--- a/controllers/apartments_controller.rb
+++ b/controllers/apartments_controller.rb
@@ -2,18 +2,30 @@ require 'pry'
 
 get '/apartments' do
   @apartments_all = Apartment.all
-  binding.pry
   erb :"apartments/index"
 end
 
-get '/apartments/:apartment_id' do
-  "ur mom in this apartment"
-end
-
 get '/apartments/new' do
-  "ur mom in a new place"
+  erb :"apartments/new"
 end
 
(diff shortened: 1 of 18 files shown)


controllers/apartments_controller.rb
```diff
@@ -0,0 +1,31 @@
+require 'pry'
+
+get '/apartments' do
+  @apartments_all = Apartment.all
```
nolds9, line 4: Nothing will change, but note that `@apartments` is convention

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+get '/apartments/new' do
+  erb :"apartments/new"
+end
+
+post '/apartments/new' do
+  @apartment = Apartment.new(params[:address], params[:monthly_rent], params[:sqft], params[:num_beds], num_baths: params[:num_baths])
+  redirect "/apartments/show/#{@apartment.id}"
+end
+
+get '/apartments/show/:id' do
```
nolds9, line 17: we normally leave "show" out of the url, so it would just be `/apartments/:id`

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+  @apartment = Apartment.find_by(id: params[:id])
+  erb :"apartments/show"
+end
+
+get '/apartments/:id/edit' do
+  @apartment = Apartment.find_by(id: params[:id])
+  erb :"apartments/edit"
+end
+
+post '/apartments/show/:id' do
```
nolds9, line 27: I think you want a `put` or `patch` route here for update

models/apartment.rb
```diff
@@ -0,0 +1,20 @@
+class Apartment < ActiveRecord::Base
+  has_many :tenants
+  def initialize(address, monthly_rent, sqft, num_beds, num_baths)
```
nolds9, line 3: We actually don't need this method since active record sets all our our attributes as `attr_accessor`s by default.

models/tenant.rb
```diff
@@ -0,0 +1,8 @@
+class Tenant < ActiveRecord::Base
+  belongs_to :artist
+  def initialize(name, age, gender)
```
nolds9, line 3: see above

nolds9:
:+1:  Good job on this! Please see inline comments for feedback.

PR #89, opened 2015-11-06
DT Solution Landlord Final

dttruong:
comfort = 3
completeness = 3

I had trouble just on the first part -- trying to load the list of Apartments on the '/apartments' page.  It was because I didn't have apartment_id on my schema.sql.  I had Erica help me.  I added the new page but did not finish the edit one.


controllers/apartments.rb
```diff
@@ -0,0 +1,22 @@
+get '/apartments' do
+	puts "*" * 30
+	@apartments_all = Apartment.all
```
nolds9, line 3: Won't change anything, but the convention is to use `@apartments`

controllers/apartments.rb
```diff
@@ -0,0 +1,22 @@
...
+	@apartment = Apartment.find(params[:id])
+	erb :"apartments/:id"
+end
+
+get '/new' do
+	puts "*" * 30
+	erb :"apartments/new"
+end
+
+post '/new' do
```
nolds9, line 18: Normally we want to post to our index, in this case `post '/apartments'`

models/apartment.rb
```diff
@@ -0,0 +1,6 @@
+class Apartment < ActiveRecord::Base
+	has_many :tenants
+	  def add_tenant(person)
```
nolds9, line 3: This method is no longer needed

nolds9:
:+1: Good job! I would encourage you to come back to this whenever you have a moment, but otherwise review inline comments for feedback. Let me know if you have any questions

PR #91, opened 2015-11-06
Britton Landlord Solution

brittonwalker:
comfort: 3
completeness: 3

I realized I forgot to create pull request, have mercy.


controllers/apartments_controller.rb
```diff
@@ -0,0 +1,20 @@
+get "/apartments" do
+  @apartments_all = Apartment.all
```
nolds9, line 2: Not going to change anything, but convention is to use `@apartments`

controllers/apartments_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  erb :"apartments/index"
+end
+
+get "/apartments/:id" do
+  @apartment = Apartment.find_by(id: params[:id])
+  erb :"apartments/:id"
+end
+
+post "/apartments" do
```
nolds9, line 11: All looks good, but can you think of a way to simplify this?

nolds9:
:+1: Great job, please see inline comments for feedback. Let me know if you have any questions.

PR #95, opened 2016-04-12
Marygriffus solution

marygriffus:
Comfort: 5
Completeness: 5


ar_exercise/exercise.rb
```diff
@@ -29,20 +29,38 @@ class Apartment < ActiveRecord::Base
...
-# Iterate over each apartment, for each apartment, display it's address and all of it's tenants
+# # Use Active record to do the following, and store the results **in a variable**
+# # example: get every tenant in the DB
+# all_tenants = Tenant.all
+#
+# # get the first tenant in the DB
+# first_tenant = Tenant.first
+#
+# # get all tenants older than 65
+# older_tenants = Tenant.where("age > ?", 65)
```
adambray, line 40: 👍

adambray, line 40: It's good to get in the habit, as you're doing, to always pass in values to AR where strings using the `?` syntax to protect against SQL injection. Happy to talk more about this if you want.

ar_exercise/exercise.rb
```diff
@@ -29,20 +29,38 @@ class Apartment < ActiveRecord::Base
...
+#
+# # get the apartment with the address "6005 Damien Corners"
+# damien_apartment = Apartment.find_by(address: "6005 Damien Corners")
+#
+# # get all tenants in that apartment
+# damien_tenants = damien_apartment.tenants
+#
+# # Use `each` and `puts` to:
+# # Display the name and ID # of every tenant
+# # all_tenants.each {|tenant| puts tenant.id.to_s + ' ' + tenant.name}
```
adambray, line 53: This is totally fine, but the 'most idiomatic' style here would be to use string interpolation to avoid having to call `.to_s`.

ar_exercise/exercise.rb
```diff
@@ -29,20 +29,38 @@ class Apartment < ActiveRecord::Base
...
+#
+# # Iterate over each apartment, for each apartment, display its address and rent price
+# all_apartments = Apartment.all
+# # all_apartments.each {|apartment| puts apartment.address + ' - ' + apartment.monthly_rent.to_s}
+#
+# # Iterate over each apartment, for each apartment, display its address and all of it's tenants
+# all_apartments.each do |apartment|
+#   puts apartment.address
+#   apartment.tenants.each{|tenant| puts "  #{tenant.name}   Age: #{tenant.age}   Gender: #{tenant.gender}"}
```
adambray, line 62: 👍

ar_exercise/exercise.rb
```diff
@@ -51,21 +69,51 @@ class Apartment < ActiveRecord::Base
 # Hint, the following methods will help: `new`, `create`, `save`, `uddate`, `destroy`
 
 # Create 3 new apartments, and save them to the DB
+# apt1 = Apartment.new(address: "123 Lollipop Lane", monthly_rent: 3000, sqft: 800, num_beds: 2, num_baths: 2)
+# apt1.save
```
adambray, line 73: You can use `create` instead of `new` to avoid having to save on a separate line.

ar_exercise/exercise.rb
```diff
@@ -51,21 +69,51 @@ class Apartment < ActiveRecord::Base
...
+# apt2 = Apartment.new(address: "4900 Elm Street", monthly_rent: 2100, sqft: 400, num_beds: 1, num_baths: 1)
+# apt2.save
+# apt3 = Apartment.new(address: "5 Backwoods Place", monthly_rent: 2000, sqft: 5000, num_beds: 22, num_baths: 4)
+# apt3.save
+
 # Create at least 9 new tenants and save them to the DB. (Make sure they belong to an apartment)
+# tenant1 = Tenant.create(name: "Pookins Magoo", age: 45, gender: "Female", apartment_id: 2)
```
adambray, line 80: You shouldn't use hard coded IDs in a script, as they are subject to change (esp if you re-seed your DB, they may not start at 1). Instead do something like this:

``` ruby
tenant1 = Tenant.create(name: "Pookins Magoo", age: 45, gender: "Female", apartment: apt1)
# or this works too
tenant1 = Tenant.create(name: "Pookins Magoo", age: 45, gender: "Female", apartment_id: apt1.id)
# or even this!
apt1.tenants.create(name: "Pookins Magoo", age: 45, gender: "Female")
```

ar_exercise/exercise.rb
```diff
@@ -51,21 +69,51 @@ class Apartment < ActiveRecord::Base
...
+# tenant10 = Tenant.create(name: "Pinon Errata", age: 74, gender: "Female", apartment_id: 11)
+
 # Note: you'll use this little bit of code as a `seeds.rb` file later on.
 
 # Birthday!
 # It's Kristin Wisoky's birthday. Find her in the DB and change her age to be 1 year older
+# kristin = Tenant.find_by(name: "Kristin Wisoky")
+# puts kristin.name, kristin.age
+# kristin.age = kristin.age + 1
+# puts kristin.name, kristin.age
```
adambray, line 98: Technically this will only update the in-memory copy of the object. We weren't specific, but in this case you'd probably want to call `.save` to save to the DB. Or you can use `.update` or `.increment`.

```
kristin.save
```

console.rb
```diff
@@ -12,6 +12,80 @@
...
 # puts "It has  #{Apartment.first.tenants.count} tenants."
 
-binding.pry
+# binding.pry
 
-puts "ignore this line, it's just here so the binding.pry above works"
+choice = nil
+while choice != 'q'
+
+  2.times {puts "*" * 70}
```
adambray, line 20: Suggest extracting something like this into a named method to clarify what it does (and since you re-use it later on)

console.rb
```diff
@@ -12,6 +12,80 @@
...
+  end
+
+  case choice
+  when 1
+    puts 'All apartments'
+    puts 'ID    | Address                  | Rent   | Sq Feet  | Beds | Baths'
+    puts '*' * 70
+    all_apartments = Apartment.all
+    all_apartments.each do |apartment|
+      print apartment.id, (' ' * (6-apartment.id.to_s.length)), '| ',
```
adambray, line 45: Clever approach to alignment! It sometimes makes sense to put this sort of code in an instance method in the Apartment class for reusability.

console.rb
```diff
@@ -12,6 +12,80 @@
...
+    end
+
+  when 3
+    puts 'Apartments listed with associated tenants'
+    all_apartments = Apartment.all
+    all_tenants = Tenant.all
+    all_apartments.each do |apartment|
+      print apartment.address
+      puts ''
+      tenants_in_apartment = all_tenants.select {|tenant| tenant.apartment_id == apartment.id}
```
adambray, line 75: Now that you have AR, you can use associations to prevent using select. (at scale, select can be 1000x slower or more, compared to using AR associations).

``` ruby
all_apartments.each do |apartment|
  print apartment.address
  puts ' '
  tenants_in_apartment = apartment.tenants
  # ...
```

console.rb
```diff
@@ -12,6 +12,80 @@
...
+      tenants_in_apartment.each {|person| print '  '; puts person.name}
+    end
+
+  when 4
+    puts "What is the tenant's name?"
+    raw_lookup = gets.chomp
+    tenant_lookup = raw_lookup.split.map(&:capitalize).join(' ')
+    apt_lookup = nil
+    all_tenants = Tenant.all
+    this_tenant = all_tenants.find {|tenant| tenant.name == tenant_lookup}
```
adambray, line 85: Again, now that we have AR, it's better to use the AR form of methods such as find (e.g. `find_by`):

``` ruby
  puts "What is the tenant's name?"
  raw_lookup = gets.chomp
  tenant_lookup = raw_lookup.split.map(&:capitalize).join(' ')

  this_tenant = Tenant.find_by(name: tenant_lookup)
  this_apartment = this_tenant.apartment

  puts "#{tenant_lookup} is renting #{this_apartment.address}"
```

adambray:
Good job!

PR #97, opened 2016-04-13
Landlord_Wes 

wesareed:
comfort_level: 4
completeness: 4.5


ar_exercise/exercise.rb
```diff
@@ -34,38 +34,108 @@ class Apartment < ActiveRecord::Base
 all_tenants = Tenant.all
 
 # get the first tenant in the DB
+first_tenant = Tenant.first or (Tenant.find(1))
+
 # get all tenants older than 65
+older_tenants = Tenant.where("age > 65")
+
 # get all apartments whose price is greater than $2300
+expensive_apartments = Apartment.where("monthly_rent > 2300", :rent)
```
adambray, line 43: The second argument here is no necessary since you're not subbing in a value into a string.

``` ruby
# this works
seniors = Tenant.where("age > 65")

# if you need a variable for the age it would look like this:
age_to_check = 50
seniors = Tenant.where("age > ?", age_to_check)
```

ar_exercise/exercise.rb
```diff
@@ -34,38 +34,108 @@ class Apartment < ActiveRecord::Base
...
+damien_corners.tenants
+
 
 # Use `each` and `puts` to:
 # Display the name and ID # of every tenant
+
+all_tenants.each do |tenant|
+  puts tenant.name
+  puts tenant.id
+end
```
adambray, line 66: 👍

ar_exercise/exercise.rb
```diff
@@ -34,38 +34,108 @@ class Apartment < ActiveRecord::Base
...
+# Hint, the following methods will help: `new`, `create`, `save`, `update`, `destroy`
 
 # Create 3 new apartments, and save them to the DB
+apartment1 = Apartment.create(address: "4128 Baldwin Crescent", monthly_rent: 7500, sqft: 2000, num_beds: 3, num_baths: 2.5)
+apartment2 = Apartment.create(address: "4055 Joshua Barney", monthly_rent: 5000, sqft: 1500, num_beds: 2, num_baths: 2)
+apartment3 = Apartment.create(address: "1211 Abraham Drive", monthly_rent: 10000, sqft: 3500, num_beds: 5, num_baths: 4)
+
+
 # Create at least 9 new tenants and save them to the DB. (Make sure they belong to an apartment)
+new_tenant1 = Tenant.create(name: "Alexander	Hardy", age: "25", gender: "Male", apartment_id: 7)
```
adambray, line 97: You shouldn't use hard coded IDs in a script, as they are subject to change (esp if you re-seed your DB, they may not start at 1). Instead do something like this:

``` ruby
Tenant.create(name: "Alex", age: 21, gender: "female", apartment: apartment1)
# or this works too
Tenant.create(name: "Alex", age: 21, gender: "female", apartment_id: apartment1.id)
# or even this!
apartment1.tenants.create(name: "Alex", age: 21, gender: "female")
```

ar_exercise/exercise.rb
```diff
@@ -34,38 +34,108 @@ class Apartment < ActiveRecord::Base
...
+
 # Note: you'll use this little bit of code as a `seeds.rb` file later on.
 
 # Birthday!
 # It's Kristin Wisoky's birthday. Find her in the DB and change her age to be 1 year older
 # Note: She's in the seed data, so she should be in your DB
+kristin_wisoky = Tenant.find_by(name: "Kristin Wisoky")
+puts kristin_wisoky.age
+kristin_wisoky.age = kristin_wisoky.age + 1
+puts kristin_wisoky.age
```
adambray, line 117: Technically this will only update the in-memory copy of the object. We weren't specific, but in this case you'd probably want to call `.save` to save to the DB. Or you can use `.update` or `.increment`.

```
kristin.save
```

ar_exercise/exercise.rb
```diff
@@ -34,38 +34,108 @@ class Apartment < ActiveRecord::Base
...
+apartment_reno.monthly_rent = apartment_reno.monthly_rent + 400
+puts apartment_reno.monthly_rent
 
 # Millenial Eviction!
 # Find all tenants who are under 30 years old
 # Delete their records from the DB
+mill_tenants = Tenants.where("age < 30")
+mill_tenants.each do |millenial|
+  millenial.destroy
+end
```
adambray, line 141: 👍

adambray:
👍

adambray:
👍

PR #99, opened 2016-04-13
zaneh_solution

zanehmw:
comfort: 4
complete: 4.5

### DIFF
--- a/ar_exercise/exercise.rb
+++ b/ar_exercise/exercise.rb
@@ -89,7 +89,7 @@ binding.pry
 # Rennovation!
 # Find the apartment "62897 Verna Walk" and update it to have an additional bedroom
 # Make sure to save the results to your database
-verna_walk = Tenant.find_by(address: "62897 Verna Walk")
+  # verna_walk = Tenant.find_by(address: "62897 Verna Walk")
 
 # Rent Adjustment!
 # Update the same apartment that you just 'rennovated'. Increase it's rent by $400
--- a/ar_exercise/exercise.rb
+++ b/ar_exercise/exercise.rb
@@ -33,14 +33,28 @@ end
 # example: get every tenant in the DB
 all_tenants = Tenant.all
 
+binding.pry
 # get the first tenant in the DB
+  #  first_tenant = all_tenants.find(0)
+
 # get all tenants older than 65
+  # seniors = Tenant.where("age > 65", :age)
+
 # get all apartments whose price is greater than $2300
+  # expensive_apartments = Apartment.where("monthly_rent > 2300", :monthly_rent)
+
 # get the apartment with the address "6005 Damien Corners"
+  # damien_corners = Apartment.find_by(address: "6005 Damien Corners")
+
 # get all tenants in that apartment
+  # damien_residents = damien_corners.tenants
 
 # Use `each` and `puts` to:
 # Display the name and ID # of every tenant
+  # all_tenants.each do |tenant|
+  # puts "Name: #{tenant.name}, Id: #{tenant.id}"
+  # end
+
 # Iterate over each apartment, for each apartment, display it's address and rent price
 # Iterate over each apartment, for each apartment, display it's address and all of it's tenants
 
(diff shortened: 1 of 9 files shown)


ar_exercise/exercise.rb
```diff
@@ -33,14 +33,28 @@ class Apartment < ActiveRecord::Base
 # example: get every tenant in the DB
 all_tenants = Tenant.all
 
+binding.pry
 # get the first tenant in the DB
+  #  first_tenant = all_tenants.find(0)
```
adambray, line 38: IDs start at 1, not zero! And in fact, there's no guarantee that the first tenant has an ID of say, 12 (if the first 11 were created and then deleted).

A better way would be:

``` ruby
Tenant.first
# or 
Tenant.all.first
```

ar_exercise/exercise.rb
```diff
@@ -33,14 +33,28 @@ class Apartment < ActiveRecord::Base
...
+
 # get all tenants older than 65
+  # seniors = Tenant.where("age > 65", :age)
```
adambray, line 41: The second argument here is no necessary since you're not subbing in a value into a string.

``` ruby
# this works
seniors = Tenant.where("age > 65")

# if you need a variable for the age it would look like this:
age_to_check = 50
seniors = Tenant.where("age > ?", age_to_check)
```

ar_exercise/exercise.rb
```diff
@@ -33,14 +33,28 @@ class Apartment < ActiveRecord::Base
...
+
 # get all apartments whose price is greater than $2300
+  # expensive_apartments = Apartment.where("monthly_rent > 2300", :monthly_rent)
```
adambray, line 44: Same as above

ar_exercise/exercise.rb
```diff
@@ -33,14 +33,28 @@ class Apartment < ActiveRecord::Base
...
+
 # get the apartment with the address "6005 Damien Corners"
+  # damien_corners = Apartment.find_by(address: "6005 Damien Corners")
+
 # get all tenants in that apartment
+  # damien_residents = damien_corners.tenants
```
adambray, line 50: 👍

ar_exercise/exercise.rb
```diff
@@ -51,16 +65,31 @@ class Apartment < ActiveRecord::Base
 # Hint, the following methods will help: `new`, `create`, `save`, `uddate`, `destroy`
 
 # Create 3 new apartments, and save them to the DB
+  # zanehs_house = Apartment.create(address: "29 Fairhill Lane", monthly_rent: 200, sqft: 3000, num_beds: 3, num_baths: 4)
+  # grandmas_house = Apartment.create(address: "6508 Springbrook Lane", monthly_rent: 400, sqft: 3200, num_beds: 4, num_baths: 6)
+  # random_house = Apartment.create(address: "670 Fairy Lane", monthly_rent: 900, sqft: 2000, num_beds: 2, num_baths: 1)
+
 # Create at least 9 new tenants and save them to the DB. (Make sure they belong to an apartment)
 # Note: you'll use this little bit of code as a `seeds.rb` file later on.
+  # Tenant.create(name: "Zaneh", age: 23, gender: "female", apartment_id: 21)
```
adambray, line 74: You shouldn't use hard coded IDs in a script, as they are subject to change (esp if you re-seed your DB, they may not start at 1). Instead do something like this:

``` ruby
Tenant.create(name: "Ajeh", age: 21, gender: "female", apartment: zanehs_house)
# or this works too
Tenant.create(name: "Ajeh", age: 21, gender: "female", apartment_id: zanehs_house.id)
# or even this!
zanehs_house.tenants.create(name: "Ajeh", age: 21, gender: "female")
```

ar_exercise/exercise.rb
```diff
@@ -51,16 +65,31 @@ class Apartment < ActiveRecord::Base
...
 
 # Birthday!
 # It's Kristin Wisoky's birthday. Find her in the DB and change her age to be 1 year older
 # Note: She's in the seed data, so she should be in your DB
+  # all_tenants.find_by(name: "Kristin Wisoky")
 
 # Rennovation!
 # Find the apartment "62897 Verna Walk" and update it to have an additional bedroom
 # Make sure to save the results to your database
+  # verna_walk = Tenant.find_by(address: "62897 Verna Walk")
```
adambray, line 92: Where's the incrementing?

console.rb
```diff
@@ -1,17 +1,49 @@
+require "bundler/setup" # require all the gems we'll be using for this app from the Gemfile. Obviates the need for `bundle exec`
+require "pg" # postgres db library
+require "active_record" # the ORM
+require "pry" # for debugging
+
 require_relative "db/connection"
-require_relative "models/apartment"
-require_relative "models/tenant"
+require_relative "models/apartments"
```
adambray, line 7: filenames for classes should be singular! (that's not a technical limitation, but a convention we try to follow as ruby devs)

db/seeds.rb
```diff
@@ -0,0 +1,29 @@
...
+Tenant.destroy_all
+Apartment.destroy_all
+
+
+
+zanehs_house = Apartment.create(address: "29 Fairhill Lane", monthly_rent: 200, sqft: 3000, num_beds: 3, num_baths: 4)
+grandmas_house = Apartment.create(address: "6508 Springbrook Lane", monthly_rent: 400, sqft: 3200, num_beds: 4, num_baths: 6)
+random_house = Apartment.create(address: "670 Fairy Lane", monthly_rent: 900, sqft: 2000, num_beds: 2, num_baths: 1)
+
+Tenant.create(name: "Zaneh", age: 23, gender: "female", apartment_id: zanehs_house[:id])
```
adambray, line 19: This works, but since the house variables hold AR Instances / objects, you can use the _dot_ syntax to call the id method instead of brackets:

```
zanehs_house.id
```

adambray:
:

console.rb
```diff
@@ -1,17 +1,49 @@
...
+#   Tenant.new(tenant[:id], tenant[:name], tenant[:age], tenant[:apartment_id])
+# end
+puts "*" * 50
+puts "Hello welcome to this app."
+puts "Type '1' to view all of your apartments"
+puts "Type '2' to view all of your tenants"
+puts "Type '3' to view all of your tenants"
+track_number = gets.chomp
+if track_number == "1"
+  puts apartments
```
adambray, line 39: This will display each apartment in a not-super helpful way. That's because `puts` is going to 'convert' the apartment to a string, and give you something like `<#Apartment:x12301238 .....>`.

Instead, iterate using each, and puts the address etc:

``` ruby
apartments.each do |apt|
  puts "Address: #{apt.address} Rent: #{apt.monthly_rent}" # etc
end
```

console.rb
```diff
@@ -1,17 +1,49 @@
...
+elsif track_number == "2"
+  puts tenants
+elsif track_number == "3"
+  #trying to figure out the code that goes here
```
adambray, line 43:
``` ruby
apartments.each do |apt|
  puts "Address: #{apt.address} Rent: #{apt.monthly_rent}" # etc
  apt.tenants.each do |tenant|
    # what goes here?
  end 
end
```

adambray:
👍

PR #108, opened 2016-04-13
Landlord problem - new commit

LOHankins:


ar_exercise/exercise.rb
```diff
@@ -26,23 +26,86 @@ class Apartment < ActiveRecord::Base
...
+puts "*" * 50
+
 all_tenants = Tenant.all
+all_apartments = Apartment.all
+all_tenants.map {|tenant| puts tenant.name + "  " + tenant.age}
+
+puts "*" * 50
 
 # get the first tenant in the DB
+first_tenant = Tenant.find(1)
```
tylercrosse, line 43: This assumes that the first tenant has an id of 1, however if this tenant was destroyed the first tenant might be 2 or, any other number. A better way of finding the first tenant is to do `Tenant.first`

ar_exercise/exercise.rb
```diff
@@ -26,23 +26,86 @@ class Apartment < ActiveRecord::Base
...
+
+puts "*" * 50
+
 # get all tenants older than 65
+all_tenants.map do |tenant| 
+	if (tenant.age.to_i > 65)
+		puts tenant.name + "  " + tenant.age
+	# puts first_tenant.name + "  " + first_tenant.age
+	end
+end
```
tylercrosse, line 55: This works, you could also have just said, which is a little shorter

``` rb
Tenant.where("age > '65' ")
```

db/landlord_schema.sql
```diff
@@ -0,0 +1,20 @@
...
+
+CREATE TABLE apartments (
+  id SERIAL PRIMARY KEY,
+  address TEXT NOT NULL,
+  monthly_rent INT,
+  sqft INT,
+  num_beds INT,
+  num_baths INT
+);
+
```
tylercrosse, line 20: 👍

ar_exercise/exercise.rb
```diff
@@ -51,6 +114,22 @@ class Apartment < ActiveRecord::Base
...
+	apt.address = gets.chomp
+	puts "enter monthly rent"
+	apt.monthly_rent = gets.chomp.to_i
+	puts "enter square feet "
+	apt.sqft = gets.chomp.to_i
+	puts "enter num beds "
+	apt.num_beds = gets.chomp.to_i
+	puts "enter num bathrooms"
+	apt.num_baths = gets.chomp.to_i
+end
```
tylercrosse, line 130: Instead of asking for user input, a much better way to do this would be to just hardcode them in.

``` rb
# using a local variable so I can do stuff with them later
apartment1 = Apartment.create(address: "1600 Pennsylvania Ave", "more key value pairs etc. etc." )
```

ar_exercise/exercise.rb
```diff
@@ -51,6 +114,22 @@ class Apartment < ActiveRecord::Base
...
+
+
```
tylercrosse, line 132: You could then use the variables for the apartments to create tenants that are attached to those apartments. ActiveRecord is super smart and will automatically add the appropriate `apartment_id` for us when we create tenants like this:

``` rb
apartment1.tenants.create("some key value pairs for tenant properties, not including apartment_id")
```

tylercrosse:
Nice job figuring out the exercises 👍 I encourage you to take a careful look at the solution branch and see how some of your solutions could have been done more cleanly using ActiveRecord methods like `where`

PR #109, opened 2016-04-14
Completed All Requirements

7ruth:
Comfort: 4
Completeness: 5

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,57 @@
+require 'pry'
+require 'active_record'
+
+## Connect to the database
+require_relative 'db/connection'
+
+# Load models
+require_relative 'models/apartment'
+require_relative 'models/tenant'
+
+require 'sinatra'
+require 'sinatra/reloader'
+
+get '/' do
+  erb :"index"
+end
+
+get '/apartments' do
+  @apartments = Apartment.all
+  erb :"apartments/index"
+end
+
+get '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/show"
+end
+
+post '/apartments/new' do
+  Apartment.create(address:params[:apartmentAddress], monthly_rent:params[:apartmentMonthlyRent], sqft:params[:apartmentSqft], num_beds:params[:apartmentNumBeds], num_baths:params[:apartmentNumBaths])
+  redirect "/apartments"
+end
+
+get '/apartments/:id/tenants' do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/tenants"
+end
+
+post '/apartments/:id/tenants/new' do
+  Tenant.create(name:params[:tenantName] , age:params[:tenantAge] , gender:params[:tenantGender], apartment_id:params[:id] )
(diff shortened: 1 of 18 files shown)


views/apartments/tenants.erb
```diff
@@ -0,0 +1,28 @@
+<script type="text/javascript">
+  $(document).ready(function(){
+      $("button").click(function(){
+        if (this.innerHTML == "Add A Tenant"){
+          $("form").hide(1000);
+          $("#create").toggle('slow','swing');
+        }
+      });
+  });
```
tylercrosse, line 9: This is a nice little snippet. It would better to create a js file for it in `public/js/coolSnippet.js` and link to it. That way you don't have to duplicate it on multiple views and you can better separate your concerns.

views/layout.erb
```diff
@@ -0,0 +1,17 @@
...
+<body>
+
+<header>
+  <a href='/'>Lord of Land</a>
+</header>
+
+ <%= yield %>
+
+</body>
+</html>
```
tylercrosse, line 17: 👍

app.rb
```diff
@@ -0,0 +1,57 @@
...
+end
+
+get '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/show"
+end
+
+post '/apartments/new' do
+  Apartment.create(address:params[:apartmentAddress], monthly_rent:params[:apartmentMonthlyRent], sqft:params[:apartmentSqft], num_beds:params[:apartmentNumBeds], num_baths:params[:apartmentNumBaths])
+  redirect "/apartments"
```
tylercrosse, line 30: A slightly cleaner way to do this is to use nested params. See the corresponding comment on `apartments/index` This would allow you to instead just say

``` rb
Apartment.create(params[:apartment])
```

models/Apartment.rb
```diff
@@ -0,0 +1,3 @@
+class Apartment < ActiveRecord::Base
```
tylercrosse, line 1: The convention is to use lowercase for the filenames, even for classes

views/apartments/index.erb
```diff
@@ -0,0 +1,33 @@
...
+</script>
+
+<!--Buttons for selection options  -->
+<button type="button">Add An Apartment</button>
+
+
+<!--Create Form Apartment-->
+<form id = "create" action = '/apartments/new' method="post" style="display:none; width:15vw;">
+  <input name='apartmentAddress' placeholder="Add Apartment Address">
+  <input name='apartmentMonthlyRent' placeholder="Add Apartment Rent (optional)">
```
tylercrosse, line 20: You can nest the names on forms, which will clean up the params that are submitted when this form gets submitted, and in turn cleans up the controller actions. Additionally, it's better to also include labels instead of just relying on placeholder text.

``` html
  <label for="apartment[sqft]">sqft</label>
  <input type='text' name='apartment[sqft]' placeholder="2000">
```

tylercrosse:
Nice job! 👍 I made a couple small notes about potential ways to clean up and refactor.

PR #113, opened 2016-04-15
Landlord Lab

mvaden:
Didn't quite finish/get my MVP. I followed the lesson plan for most things. It felt sort of comfortable. Couldn't get my we page to render beyond the home page. Lost a lot of time dealing with syntax errors for "require" and "require relative" for much of the beginning. Gained some comfort otherwise.

Comfort Level: 2.5
Completed: 3


app.rb
```diff
@@ -0,0 +1,74 @@
...
+delete '/apartments/:address' do
+  @apartments = Apartment.find(params[:address])
+  @apartments.destroy
+  redirect("/apartments")
+end
+
+# edit apartments
+get "/apartments/:address/edit" do
+  @apartments = Apartment.find(params[:address])
+  erb :"apartmentss/edit"
```
adambray, line 35: looks like a typo here!

app.rb
```diff
@@ -0,0 +1,74 @@
...
+end
+
+# update apartements
+put '/apartments/:address' do
```
adambray, line 39: Usually we'll use ID here, since that's the simplest unique thing that identifies an apartment. Also, addresses have spaces and other characters in that aren't fit for use in a URL.

adambray:
One note, it looks like you had a lot of code (views and 'controllers') here that didn't work. Instead of adding a lot of code that doesn't work, focus on adding 1 feature at a time (a small amount of code at a time) and working on that until it works.

PR #114, opened 2016-04-15
Marygriffus solution

marygriffus:
Comfort: 5
Completeness: 5


app.rb
```diff
@@ -0,0 +1,111 @@
...
+end
+
+
+get '/tenants/new' do
+  @tenants = Tenant.all
+  @apartments = Apartment.all
+  erb :"/tenants/new"
+end
+
+post '/new_tenant' do
```
adambray, line 63: customarily this route would be `post /tenants`, but this works.

app.rb
```diff
@@ -0,0 +1,111 @@
...
+  @tenant = Tenant.find(params[:id])
+  @apartment = @tenant.apartment
+  @roommates = @apartment.tenants - [@tenant]
+  erb :"/tenants/delete"
+end
+
+delete '/tenants/:id' do
+  @tenant = Tenant.find(params[:id])
+  @apartment = @tenant.apartment
+  @roommates = @apartment.tenants - [@tenant]
```
adambray, line 92: Is there a reason you had to do this? Destroying a tenant should remove them from the apartment since a tenant belongs_to an apartment. In other words, the relationship is stored on the tenant, so deleting the tenant deletes the relationship.

adambray:
👍

PR #115, opened 2016-04-15
Landlord

JoshNezi:
Level of Completeness 1
Level of Comfort 4 

I couldn't move forward with this assignment because of a database error.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,25 @@
+require "pg"
+require "bundler/setup"
+require "active_record"
+require "pry"
+
+require 'sinatra'
+require 'sinatra/reloader'
+require 'active_record'
+# Load the file to connect to the DB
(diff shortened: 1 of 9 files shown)


app.rb
```diff
@@ -0,0 +1,25 @@
...
+
+get '/apartments' do
+  @apartments = Apartment.all
+  erb :"apartments/index"
+end
+
+get '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/show"
+end
```
tylercrosse, line 25: This looks like a great start 👍

tylercrosse:
Hopefully the database error was the one we paired on. In the future the more specfic you are about the errors and problems you are having the better we'll be at giving you feedback. Seems like you were on the right track though 👍

PR #117, opened 2016-04-15
Landlord HW - AR, ERB

lelandjordon:
Completeness: 3
Comfort: 3

Same comfort level as Tunr because I copy-pasted a lot of stuff over from Tunr.  The abstraction of Ruby and AR doing things behind the scenes is getting confusing.


views/show.erb
```diff
@@ -0,0 +1,11 @@
+<!DOCTYPE html>
```
tylercrosse, line 1: This file isn't being used and could be deleted

views/index.erb
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
tylercrosse, line 1: If you wanted to keep a root view it probably would be better to rename it `root` instead. Another alternative would be redirect users visiting `'/'` to another existing view like `apartments/index`

views/layout.erb
```diff
@@ -0,0 +1,14 @@
...
+  <head>
+    <meta charset="utf-8">
+    <title>Layout</title>
+    
+  </head>
+  <body>
+    <h1>This Land is MY LAND</h1>
+    <h2><a href="/apartments">All Apartments</a></h2>
+    <h3><a href="/tenants">All Tenants</a></h3>
+    <%= yield %>
```
tylercrosse, line 12: Sinatra places all of the other views inside of this `<%= yield %>`. This means that you no longer need to have the `<!DOCTYPE html> <html> yada yada` stuff in any of the other views

app.rb
```diff
@@ -0,0 +1,59 @@
...
+require "pry"
+
+require_relative "models/apartment"
+require_relative "models/tenant"
+require_relative "db/connection"
+
+get '/' do
+  "Home page"
+  erb :index
+end
```
tylercrosse, line 12: This could also simply redirect users to an existing view. You also don't need the "Home Page"

app.rb
```diff
@@ -0,0 +1,59 @@
...
+
+get '/apartments' do
+  "Hello Apartments"
```
tylercrosse, line 15: This line is not needed

views/tenants/show.erb
```diff
@@ -0,0 +1,22 @@
...
+   <meta charset="utf-8">
+   <title><% @tenant.name %> </title>
+ </head>
+ <body>
+   <h1>Hello World - this is views/tenants/show.erb</h1>
+ <p>
+   <%= "Name: " + @tenant.name + " Age: " + @tenant.age.to_s + " Gender: " + @tenant.gender %>
+
+ </p>
+ <h3>DISPOSE of this tenant!!</h3>
```
tylercrosse, line 13: 😆

tylercrosse:
Really nice job overall! 👍 The biggest thing I noticed was the repeated html boilerplate code on every view. Since you are using a layout this isn't needed.

PR #119, opened 2016-07-23
Tyler Rosenberg - Landlord WITH AR Solution

TyRyanR:
Comfort: 4
Completeness: 4

At first had a little trouble setting up the connection and all required files because I was off by a SINGLE letter. Had trouble with the seed.rb file. For some reason, the apartment ID's kept being created starting at number 30. So for my three apartments, whenever I ran the seed.rb file, the apartment Id's would be 30, 31, and 32.. That messed up the connection between my tenants' apartment_id foreign attribute values and the apartment Id's.

### DIFF
--- a/ar_exercise/exercise.rb
+++ b/ar_exercise/exercise.rb
@@ -5,19 +5,6 @@ require "pg" # postgres db library
 require "active_record" # the ORM
 require "pry" # for debugging
 
-ActiveRecord::Base.establish_connection(
-  :adapter => "postgresql",
-  :database => "landlord"
-)
-
-class Tenant < ActiveRecord::Base
-  belongs_to :apartment
-end
-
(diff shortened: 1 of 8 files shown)


ar_exercise/exercise.rb
```diff
@@ -5,67 +5,109 @@
...
+kristin = Tenant.find_by(name: "Kristin Wisoky")
+kristin.age += 1
+
 # Rennovation!
 # Find the apartment "62897 Verna Walk" and update it to have an additional bedroom
 # Make sure to save the results to your database
 
+apartment = Apartment.find_by(address: "62897 Verna Walk")
+aparment.num_beds += 1
+
```
nayana487, line 93: nice, you got a good handle on your AR queries.

PR #122, opened 2016-07-24
Jac Ahn Landlord w AR hw

jacahn:
Comfort: 2
Completeness: 4

This was a really excellent exercise that I struggled mightily with. I leaned heavily on other students' code to understand how to do stuff (Thanks Megan, Tyler and Evan!) I enjoyed it very much.

Here were my issues:
1. This was the answer I saw others had to find tenants younger than 20, yet I got the answer below. Don't know why.
    `[4] pry(main)> Tenant.where("age: < ?", 20)
    => #<Tenant::ActiveRecord_Relation:0x3fe552d2c770>`
2. At the end of Step 2, when running console.rb in bash:
   
   ``` Jacs-MacBook-Pro:landlord jacahn$ ruby console.rb
   There are 3 apartments
   
   There are 9 tenants
   ---
   
   The first apartment is at 111 Huey Way.
   
   It has  0 tenants.

```
Every time I reload console.rb, the apartments are assigned new indexes, so what I hard coded as tenants’ apartments are not remaining their apts. How do I fix this?

3. For Step 3 of the Active Record exercises, I could not get the third request of the CLI to work:
# See a list of all apartments and their associated tenants (just address and name)

I don't know the right code, but this is what I wrote:

```

def all_apts_tenants
  puts "List of Apartments and Tenants"
  Apartment.all.each do |apartment|
    puts "#{apartment.address} and #{apartment.tenants.inspect}"
  end
end

```
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,5 @@
+source 'https://rubygems.org'
+
+gem "pg"  # this gem allows ruby to talk to postgres
+gem "activerecord"  # this gem provides a connection between your ruby classes to relational database tables
+gem "pry"  # this gem allows access to REPL
--- a/Gemfile
+++ b/Gemfile
@@ -3,3 +3,5 @@ source 'https://rubygems.org'
 gem "pg"  # this gem allows ruby to talk to postgres
 gem "activerecord"  # this gem provides a connection between your ruby classes to relational database tables
 gem "pry"  # this gem allows access to REPL
+gem "sinatra"
+gem "sinatra-contrib"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,40 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    activemodel (5.0.0)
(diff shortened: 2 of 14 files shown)


jacahn:
Apologies for the issues with the " ``` "s. LOL

jacahn:
Landlord with Sinatra & AR

comfort:4
completeness:4

This was a very good exercise to learn the basics of Sinatra and AR.

andrewsunglaekim:
1. This is using active record to write a part of a raw SQL query.
2.  This is because if you indexes don't reset at 0. So you'd have to drop the database and recreate. Alternatively when seeding the tenants pass in actual apartments rather than integers.
3. For each apartment you should also iterate through the tenants. Something like this:

``` ruby
@apartments.each do |apartment|
  puts apatment.address
  apartment.tenants.each do |tenant|
    puts tenant.name
  end
end
```

andrewsunglaekim:
Good job on this exercise! 👍

PR #124, opened 2016-07-25
hw

pamelaolney:
Comfort: 3
Completeness: 4

I am at a point where I tried my best.  This is really a lot and questions my motives to become a developer.  I really didnt understand why I couldn't complete Step 3.  If I could just write a simple 

`if response == 1 do
  puts Apartment.all
end`

Then it would make some sense to me.  I was really confused on that part.  Was incredibly sick with a high fever which cut my time on being able to complete this hw.  Def, tried my best.  Not feeling good about this so far.

### DIFF
--- a/console.rb
+++ b/console.rb
@@ -1,3 +1,5 @@
+require "pry"
+
 require_relative "db/connection"
 require_relative "models/apartment"
 require_relative "models/tenant"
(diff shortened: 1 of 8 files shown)


console.rb
```diff
@@ -12,6 +14,14 @@
...
+# puts "ignore this line, it's just here so the binding.pry above works"
+puts "Please select from the following list of options"
+puts "1) View all Apartments"
+puts "2) Veiw all Tenants"
+puts "3) View all Apartments and there associated Tenants"
 
-puts "ignore this line, it's just here so the binding.pry above works"
+response = gets.chomp
+#I'm really stuck here guys.  Not sure where to move forward.  This is the work that I have completed.
+# Everything I have tried is not working.  I guess I'll just meet up with someone during office hours to complete this part.
```
superbuggy, line 27: We can def go over this during our 1-on-1 this week, but I think the code you had in your PR request should work! Good effort and progress otherwise.

superbuggy:
Like I said in file comments, we can go over this assignment during our 1-on-1 👍. Reasonable amount of progress made, tho!

PR #125, opened 2016-07-25
Dani HW

danielanoir:
completeness: 1
comfort: 2

comments: I couldn't get the first part to work. I worked on the other assignment much longer than I expected so I didn't save enough time to focus on this. I will continue tomorrow.

### DIFF
--- a/ar_exercise/exercise.rb
+++ b/ar_exercise/exercise.rb
@@ -5,18 +5,18 @@ require "pg" # postgres db library
 require "active_record" # the ORM
 require "pry" # for debugging
 
-ActiveRecord::Base.establish_connection(
-  :adapter => "postgresql",
-  :database => "landlord"
-)
-
-class Tenant < ActiveRecord::Base
-  belongs_to :apartment
-end
-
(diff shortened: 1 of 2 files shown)


ar_exercise/exercise.rb
```diff
@@ -34,9 +34,17 @@ class Apartment < ActiveRecord::Base
 all_tenants = Tenant.all
```
superbuggy, line 34: See the note-- don't comment out that stuff-- it will prevent ActiveRecord from loading

ar_exercise/exercise.rb
```diff
@@ -34,9 +34,17 @@ class Apartment < ActiveRecord::Base
...
```
superbuggy, line 35: All the methods below are from ActiveRecord `.find` `.where` `.find_by`; these will not work since you've commented outlines 8-19. What've you written is correct, though. You would just have no way of knowing or testing if you comment out the line where the models inherit form ActiveRecord::Base

superbuggy:
👍

PR #127, opened 2016-07-25
Aubrey Northam's Landlord (with AR) HW

aubreynortham:
Comfort: 3.5
Completeness: 5

The file connectivity is what trips me up here. The syntax and everything about the commands makes sense, and they are getting more repetitive now so I'm definitely feeling ok there. But the way all the files relate to each other always slows me down, along with getting the terminal commands just right in order to test the code I want. But I got through this one ok, it just took longer than I wanted!

### DIFF
--- a/ar_exercise/exercise.rb
+++ b/ar_exercise/exercise.rb
@@ -1,5 +1,4 @@
-### NOTE: Make sure you've loaded the seeds.sql file into your DB before starting
-### this exercise
+
 
 require "pg" # postgres db library
 require "active_record" # the ORM
@@ -19,7 +18,7 @@ class Apartment < ActiveRecord::Base
 end
 
 ################################################
-#### NOTE: DON'T MODIFY ABOVE THIS LINE     ####
+#: DON'T MODIFY ABOVE THIS LINE     ####
 ################################################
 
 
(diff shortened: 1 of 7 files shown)


ar_exercise/exercise.rb
```diff
@@ -31,18 +30,31 @@ class Apartment < ActiveRecord::Base
...
+# damien_corners_apt = Apartment.find_by(address: "6005 Damien Corners")
 # get all tenants in that apartment
-
+# damien_corners_apt.tenants
 # Use `each` and `puts` to:
 # Display the name and ID # of every tenant
+# Tenant.all.each do |tenant|
+#   puts "#{tenant.name}: #{tenant.id}"
+# end
 # Iterate over each apartment, for each apartment, display it's address and rent price
```
superbuggy, line 50: 👍  on all above

console.rb
```diff
@@ -2,16 +2,47 @@
...
+        puts "Address: #{apartment.address}, tenants include #{tenant.name}"
+      end
+    end
+  else
+    puts "Goodbye"
+  end
+end
+
+
+
```
superbuggy, line 45: 👍

db/connection.rb
```diff
@@ -0,0 +1,8 @@
+require "pg" # postgres db library
+require "active_record" # the ORM
+require "pry" # for debugging
+
```
superbuggy, line 4: these are requiring gems, rather than files-- `require_relative` is for files

PR #129, opened 2016-07-25
Nelley's LandlordAR

mrnelley:
comfort: 4.5
completeness: 4.5

I was surprisingly more knowledgeable about what I was doing than I thought. Now to work on this confidence thing... I think if i were to be more sure of myself as a coder who knows the syntax and structure for what software should look like, then I would undoubtedly code faster.


superbuggy:
👍 confidence will come with reps, i promise! isn't AR great tho?

PR #131, opened 2016-07-25
Ahmed's HW 07/25

ahmedelk91:
Comfort: 3

Completeness: 2.5

Comments: I had a lot of difficulty with this one, even though I felt like I knew what I was doing and understood why I was doing it. I was unable to run the ruby files due to an error with the json gem. After hours of google fu, I decided to time box myself with this one. It was proving to be too difficult. I am looking to fully complete this on my own, but would like help on that technical issue so I can proceed. Thank you!


console.rb
```diff
@@ -1,10 +1,15 @@
+require "bundler/setup"
+require "pg"
+require "active_record"
+require "pry"
+
 require_relative "db/connection"
-require_relative "models/apartment"
-require_relative "models/tenant"
+require_relative "models/apartments"
+require_relative "models/tenants"
```
superbuggy, line 9: It's possible the json error resulted from pluralizing the model-- the convention is that the model name is singular ie `apartment.rb` and `tenant.rb`

superbuggy:
see if change your model files to singular fixes the problem

PR #133, opened 2016-07-25
Sarah Shultis Landlord

skshultis:
comfort:4/5
complete: 4/5

Wasn't sure if the Sinatra was supposed to be included. My app works but as of last night none of the tenants I created matched up with the properties I made? This was fun and I liked playing with it.

### DIFF
--- a/lapp.rb
+++ b/lapp.rb
@@ -1,0 +1,87 @@
+require 'sinatra'
+require 'sinatra/reloader'
+require 'active_record'
+
+# Load the file to connect to the DB
+require_relative 'db/connection'
+
+# Load models
+require_relative 'models/apartment'
+require_relative 'models/tenant'
+
+
+get '/' do
+  erb :'master'
+end
+
+get '/apartments' do
+  @apartments = Apartment.all
+  erb :"apartments/index"
+end
+
+get '/apartments/new' do
+  erb :"apartments/new"
+end
+
+get '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  erb :"apartments/show"
+end
+
+post '/apartments' do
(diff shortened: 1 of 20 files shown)


lapp.rb
```diff
@@ -0,0 +1,87 @@
...
+  @tenant = Tenant.find(params[:id])
+  @tenant.update(params[:tenant])
+  redirect "tenants/#{@tenant.id}"
+end
+
+delete '/tenants/:id' do
+  tenant = Tenant.find(params[:id])
+  tenant.destroy
+  redirect '/tenant'
+end
```
nayana487: nice job setting up your routes!

nayana487, line 87: So you'd connect the properties to the tenants here... something like the below for the index view, for example.

get '/apartments/:id/tenants' do
  @apartment = Apartment.find(params[:id])
  erb :"tenants/index"
end

views/apartments/show.erb
```diff
@@ -0,0 +1,9 @@
+<a href="/apartments/new">create new property</a>
+
+<h2>apartment address: <%=@apartment.address%></h2>
+<h3>monthly rent: <%=@apartment.monthly_rent%></h3>
+<h3>square footage: <%=@apartment.sqft%></h3>
+<h3>bedrooms: <%=@apartment.num_beds%></h3>
+<h3>bathrooms: <%=@apartment.num_baths%></h3>
+
```
nayana487, line 8: could add a delete here.

views/master.erb
```diff
@@ -0,0 +1,2 @@
+<a href="/apartments">Apartments</a>
+<a href="/tenants">Tenants</a>
```
nayana487, line 2: could have also had a redirect to the apartments index

PR #135, opened 2016-07-26
Tyler Rosenberg - July 28 HW - Landlord Sinatra AND Active Record

TyRyanR:
Comfort: 3.8
Completeness: 4

Definitely getting more used to working with all these files the more exercises I do. Still a little confused on put and delete paths.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -15,3 +15,8 @@ get '/apartments' do
   @apartments = Apartment.all
   erb :'apartments/index'
 end
+
+get '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  erb :'apartments/show'
+end
--- a/app.rb
+++ b/app.rb
@@ -29,3 +29,15 @@ post '/apartments' do
   @apartment = Apartment.create(params[:apartment])
   redirect "/apartments/#{@apartment.id}"
 end
+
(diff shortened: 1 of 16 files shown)


app.rb
```diff
@@ -0,0 +1,43 @@
...
+  @apartment = Apartment.find(params[:id])
+  erb :'apartments/edit'
+end
+
+
+put '/apartments/:id' do
+  @apartment = Apartment.find(params[:id])
+  @apartment.update(params[:apartment])
+  redirect "/apartments/#{@apartment.id}"
+end
```
nayana487, line 43: nice job setting up your routes!

ar_exercise/exercise.rb
```diff
@@ -5,67 +5,109 @@
...
+
+less_than_30 = Tenant.where("age  < ?", 30)
+
+less_than_30.each |tenant|
+  tenant.destroy
+end
+
+
+binding.pry
+
```
nayana487, line 112: you got a good handle on using AR queries.

nayana487:
Hey Tyler, nice job. Looks like you got a good handle on the material, but you have some tenant views missing. Are you confused about the update and delete methods because they don't correspond with a view? You would just have a redirect to another page in the controller.

PR #136, opened 2016-07-26
Megan's HW - Sinatra

megancoyle:
Comfort level: 4
Completeness: 5

It's going to take me a little while to get familiar with setting up these different files. It's hard for me to do that now without referencing previous assignments. Hope to get to the point where I don't have to do that very much.


andrewsunglaekim:
so pretty! Its definitely going to feel that way for a bit. I think bigger picture is to understand why the files are doing what they're doing. Using existing applications for reference is par for the course for development in general! Great work. 👍

PR #144, opened 2016-07-28
Rasheeda A. HW

RasheedaAlexander:
Comfort: 4
Completeness: 5

I kept running into a pry error with my console.rb file. But I will correct it during the lunch break.


andrewsunglaekim:
👍  This looks great Rasheeda, let me know if you have any questions.

PR #162, opened 2016-10-06
Landlord HW - Latoya Watson

LWatsonlm:
Comfort: 4/5
Completeness: 2.5-3/5 

Wasn't able to complete due to spending almost 3 hours on Wednesday working with Nick and Adrian through weird SSL cert issues in Ruby, pry not being able to run without Readline issues, and Postgres needing to be install and reinstall.  It was lame.


jshawl:
👍 But the lameness is gone now, right?!

PR #174, opened 2016-11-03
Juliet Berman Landlord

julietberman:
Comfort: 5
Completion: 5

Doing all the "reps" definitely helps!


andrewsunglaekim:
👍  awesome! that's definitely the idea.

PR #184, opened 2016-11-03
Kevin Herrera's Landlord hw

herrerake:
comfort: 2.9
completeness: 3.0
comments: Did not get to complete this assignment. I will return to is, but for now will focus on getting reps on ruby tutorial


superbuggy:
looks like you got most of it done 👍 overall

PR #186, opened 2017-02-15
Jared Edwards HW Landlord

jarededwards:


jarededwards:
Completness 4

Comfort level 3-4
I still need some work with Active Record.

nayana487:
👍

PR #192, opened 2017-02-15
Lisa Burr

laburr27:
Confidence: Thought it was 4 but now feeling 2.75-ish
Completeness: Made it mostly though exercise.rb.

After being in Sinatra I found it difficult to downshift to AR.  I understand the concepts involved I just kept getting messed up in PRY (SHOOT ME NOW!!!).  I am definitely much more comfortable working in Sinatra and am looking forward to Rails.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,5 @@
+source "https://rubygems.org"
+
+require "pg"
+require "active_record"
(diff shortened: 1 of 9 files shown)


superbuggy:
👍

PR #193, opened 2017-02-15
Landlord HW (Justin Shin)

Justin-Shin:
Completion: 2.5
Comfort Level: 5

I'm comfortable with the material, but I'm sick of landlord.  I have absolutely no motivation to really pursue this anymore.  We already had to program this and refactor it previously, and made to essentially remake it and then refactor it again into sinatra is frustrating to say the least.  I get that developers are asked to do things they don't like, but I haven't seen the developers I work with have to reinvent the wheel 4 times in such a short period of time.  This sort of exercise makes me actively dislike coding, and at this point I'm doing it for pure completion.  Even more, when an assignment has you go one direction (once again, my inclination was to make a site using sinatra but went with the guided portion instead) and then tells you to do what you wanted to do in the first place after making you do other work- makes me hate it.

On top of that, I didn't realize we would sinatraize it, so halfway through the CLI portion, I saw it and didn't even want to do the CLI part.  What's the point when we're just going to remake it?

Didn't do Sinatra because my car's headlights broke and I've been trying to fix it when I got home.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+source "https://rubygems.org"
+
+gem "pg"
+gem "activerecord"
+gem "pry"
+
+# gem "rails"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,0 +1,40 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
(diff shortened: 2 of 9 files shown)


superbuggy:
👍

PR #195, opened 2017-02-15
Clark

nrc0004:


ar_exercise/exercise.rb
```diff
@@ -34,14 +34,31 @@ class Apartment < ActiveRecord::Base
...
+SELECT * FROM apartments WHERE monthly_rent >= 2300;
 # get the apartment with the address "6005 Damien Corners"
+SELECT * FROM apartments WHERE address = '6005 Damien Corners';
 # get all tenants in that apartment
+SELECT * FROM tenants WHERE apartment_id = 6;
 
 # Use `each` and `puts` to:
 # Display the name and ID # of every tenant
+
+idName = SELECT name, tenant_id FROM tenants;
```
amaseda, line 50: Combining SQL and Ruby here won't work. Ideally you would use Active Record commands to retrieve what you need from the database.

db/seeds.rb
```diff
@@ -0,0 +1,67 @@
...
+Tenant.new(name: "Lady", age: 50, gender: "female")
+
+class Menu
+  def self.display
+    while 1
+      puts "Choose one of the following:"
+      puts "1 - See all apartemnts"
+      puts "2 - See all tenants 2"
+      input = gets.chomp
+      if (input = 1)
```
amaseda, line 32: Don't want to use a single `=` (i.e., assignment) in a conditional. Need `==` instead.

db/seeds.rb
```diff
@@ -0,0 +1,67 @@
...
+        Apartments.all
```
amaseda, line 33: Make sure you use the correct class name: `Apartment` (capitalized, singular).

db/seeds.rb
```diff
@@ -0,0 +1,67 @@
...
+      if (input = 2)
+        Tenants.all
```
amaseda, line 35: Ditto my note about `Apartment`

models/apartment.rb
```diff
@@ -0,0 +1,6 @@
+class Apartment < ActiveRecord::Base
+
+has_many :tenants
+belongs_to :tenants
```
amaseda, line 4: Should an apartment belong to a tenant? A one-to-many relationship between apartments and tenants works fine for this assignment.

Also, `belongs_to` should accept a singular argument. It doesn't apply here, but if you were to use it for these models you would write `belongs_to :tenant`.

PR #198, opened 2017-02-16
Nat Carter | landlord

ghost:
Comfort: 4.6 / 5
Completeness: 4.9 / 5

That was quite a marathon, but I needed the Active Record and Sinatra practice.

Partial or total CRUD for tenants would be nice, but for now I need to move on to the Rails intro.


ar_exercise/exercise.rb
```diff
@@ -34,38 +35,134 @@ class Apartment < ActiveRecord::Base
 all_tenants = Tenant.all
 
 # get the first tenant in the DB
+first_tenant = Tenant.first
+
 # get all tenants older than 65
+older_tenants = Tenant.where('age > 65')
```
amaseda, line 41: Take a look at this post for how to use a conditional in a `.where` statement: http://stackoverflow.com/questions/11317662/rails-using-greater-than-less-than-with-a-where-statement

PR #205, opened 2017-02-16
Savraj Singh

SavrajSingh:
complete 3
sorry adrian I got complacent and i am tired so I dont think i will complete it all
I should have started it over the weekend. it was a lot of work

comfort 2

### DIFF
--- a/ar_exercise/exercise.rb
+++ b/ar_exercise/exercise.rb
@@ -34,14 +34,21 @@ end
 all_tenants = Tenant.all
 
 # get the first tenant in the DB
+first = Tenant.first
 # get all tenants older than 65
+old = Tenant.where "age>65"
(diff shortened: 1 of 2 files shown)


ar_exercise/exercise.rb
```diff
@@ -34,14 +34,21 @@ class Apartment < ActiveRecord::Base
 all_tenants = Tenant.all
 
 # get the first tenant in the DB
+first = Tenant.first
 # get all tenants older than 65
+old = Tenant.where "age>65"
```
amaseda, line 39: You're going to need a little more here. Take a look at this post: http://stackoverflow.com/questions/11317662/rails-using-greater-than-less-than-with-a-where-statement

PR #209, opened 2017-02-16
Nyketha's Landlord AR homework

Nykke:
completeness: 5
confidence: 4
comment: this was good practice, I'm going to admit the bonus sections this weekend to get that extra practice in with sinatra.


ar_exercise/exercise.rb
```diff
@@ -28,21 +28,45 @@ class Apartment < ActiveRecord::Base
...
+end
+
 # Iterate over each apartment, for each apartment, display it's address and rent price
+apartments.each do |apt|
+  puts "#{apt.address} #{apt.monthly_rent}"
+end
+
 # Iterate over each apartment, for each apartment, display it's address and all of it's tenants
+apartments.each do |apt|
+  puts "#{apt.address}".join
```
superbuggy, line 65: don't need `.join` here

superbuggy:
I'm glad to see your comfort and proficiency increasing with your Ruby code.

 If we weren't in the middle of Rails week, I might suggest going back to do the Sinatra portion, but I would much rather recommend just focussing on the Rails material--it's more important.

PR #212, opened 2017-02-16
Landlord - Kevin Keller

kkeller93:
Comfort: 4
Completeness: 4.5

Was fine with everything up until the sinatra part. Got really confused there with all the different files that had to be navigated and populated so I missed a lot of things when initially creating the files that did not allow the code to work properly. Felt like I was banging my head against the wall for a while.


phantomhaircuts:
👍

PR #213, opened 2017-02-16
Ben Shapiro - Landlord HW

pretorben:
Completeness: 3
Comfort Level: 4

I got stuck for a long time on the seeds.rb file and once I figured out what was wrong, it was too late to keep going on it.  I should have started earlier.


superbuggy:
👍

PR #214, opened 2017-02-16
Landlord HW - Erica Norby

ericanorby:
Completeness: 4
Comfort level: 4

Fell asleep while working on this last night.. oops. I will work on it more if I have time before class.


phantomhaircuts:
👍

PR #225, opened 2017-04-15
WPurdum_Landlord

wjpurdum:
Completion - 4.5/5
Comfort level - 4/5

Specific issues:
In my console.rb, the Apartment.select("address, monthly_rent") worked fine in PRY, but not when I just ran Ruby on the file. 

I'm still a bit unsure of how we show info that is linked between 2 tables (i.e. showing the tenants for each apartment. I think we have to:
1. Set Apartment.all to a variable (i.e. all_apartments)
2. loop through that variable to get its tenants : 
all_apartments.each do |apartment| puts apartment.tenants
Is there a better way to do this? 
3. Sometimes PRY stopped coming up as an option when I rubied my file. I can't tell why this is happening - maybe because the code is stuck in a loop so binding.pry can't load?

Thanks!!

### DIFF
--- a/console.rb
+++ b/console.rb
@@ -4,13 +4,11 @@ require_relative "models/tenant"
 
 # NOTE: Uncomment the lines below to verify that your seed script is working
 
-# puts "There are #{Apartment.count} apartments"
-# puts "There are #{Tenant.count} tenants"
-#
-# puts "*" * 50
-#
-# puts "The first apartment is at #{Apartment.first.address}."
-# puts "It has  #{Apartment.first.tenants.count} tenants."
+puts "There are #{Apartment.count} apartments"
+puts "There are #{Tenant.count} tenants"
+puts "*" * 50
+puts "The first apartment is at #{Apartment.first.address}."
+puts "It has  #{Apartment.first.tenants.count} tenants."
 
 binding.pry
 
--- a/console.rb
+++ b/console.rb
@@ -3,12 +3,42 @@ require_relative "models/apartment"
 require_relative "models/tenant"
 
 # NOTE: Uncomment the lines below to verify that your seed script is working
+#
+# puts "There are #{Apartment.count} apartments"
+# puts "There are #{Tenant.count} tenants"
+# puts "*" * 50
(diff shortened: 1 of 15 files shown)


console.rb
```diff
@@ -3,15 +3,43 @@
...
+  elsif
+    response == 2
+    puts Tenant.select('name, address')
+  elsif
+    response == 3
+    all_apartments = Apartment.all
+    all_apartments.each do
+      puts all_apartments.tenants
+    end
+    end
```
juancgarcia, line 30: There's an extra `end` keyword here that doesn't match with any block. This is creating a syntax error which is why running `ruby console.rb` was failing. One of these needs to be deleted.

console.rb
```diff
@@ -3,15 +3,43 @@
...
-#
 # puts "The first apartment is at #{Apartment.first.address}."
 # puts "It has  #{Apartment.first.tenants.count} tenants."
 
+value = 0
+while value < 5 do
+puts "Welcome, landlord! To see a list of all your apartments, press 1. \n
+To see a list of all your tenants, press 2.\n
+To see a list of apartments and their tenants, press 3"
+response = gets.chomp.to_i
```
juancgarcia, line 18: You started your `while` loop expecting the input variable to be `value`, but then switched to `response` here. This technically works, but typing a number assigns  `response` only and `value` never gets changed. This means the the loop runs forever because `value` is always 0. The simplest fix is to change `value` to `response`:
```ruby
response = 0
while response < 5 do
```

juancgarcia (commented):
👍 

The reason your `console.rb` didn't work was due to an extra `end` confusing ruby and causing a syntax error.

For linked models, the way you showed is typically the approach we want to take when working with related objects. With one tiny tweak:
```ruby
all_apartments.each do |apartment|
    # putting apartments.tenants won't actually print something readable.
    # we need to access the the objects in that array and their string properties more deliberately
    # puts apartment.tenants
    apartment.tenants.each do |tenant|
        puts "#{tenant.name} - #{apartment.address}"
    end
end
```

PR #228, opened 2017-04-16
The landlord is going bankrupt...

ghost:
Comfort: 3
Completion: 3

Feeling sad, I didn't learned the core concepts.


AndyWhitley:
Really great work, you got through the majority of the exercise and your code looks mostly correct. The big think I noticed is that, in your Sinatra controllers, you want to be querying the ActiveRecord models, not using data defined in the file itself:

```rb
get "/apartments" do
  @apts = Apartment.all
  erb :index1
end
``` 

👍

PR #229, opened 2017-04-17
landlord_AR HW

StormyRamsey:
Comfort Level 2
Completion Level 2

Is Thursday not the due date for homework anymore? Feeling a little overwhelmed with the change in due dates.


juancgarcia (commented):
👍 You were on the right track, but there's still quite a bit of work left to do. Please try to continue working to finish this assignment if you can as it will serve as good practice.

PR #230, opened 2017-04-17
Curtis Ostle - Landlord 

Costle784:
completeness - 3
comfortability - 3

Still working on this...   I found the directions to be frequently unclear.


juancgarcia (commented):
👍 Nice work.

Looks like you got through most of the assignment save for the CLI and Sinatra parts. You can always come back to work on those sections if you find you'd like some more practice.

PR #231, opened 2017-04-17
Meag Doherty - Landlord AR Homework

meagdoh:
•	Comfort score, 3 out of 5
	•	Completeness score, 3 out of 5

Completed Schema and AR Exercises. Ran into trouble making the connection with db after running seeds.db

Latest ERROR message:
$ ruby console.rb 
WARN: Unresolved specs during Gem::Specification.reset:
      minitest (~> 5.1)
WARN: Clearing out unresolved specs.
Please report a bug if this causes problems.
There are 0 apartments
There are 0 tenants
**************************************************
console.rb:12:in `<main>': undefined method `address' for nil:NilClass (NoMethodError)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+source "https://rubygems.org"
+
+gem "pg"  # this gem allows Ruby to talk to postgres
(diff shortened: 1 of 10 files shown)


jsm13:
👍

PR #233, opened 2017-04-17
landlord Jeong Hoon Koo

jkoo87:
Comfort level: 2.5
Completeness: 3

While I was doing this homework, I had to spend few hours to fix a 'minitest' gem related error. There were two different versions of 'minitest' installed in my local system that I guess the program did not know which versions to use.. 
I am still working on Sinatra Views and Templates.


juancgarcia (commented):
👍 Good job.

I see you got through most of it and didn't get to finish the Sinatra part. Remember you can always come back to this if you'd like to review or get more practice.

PR #235, opened 2017-04-17
landlord

zonekid:
difficulty -3.5
comfortable-3

The directions were kind of simple.  Just a lot of work.


juancgarcia (commented):
👍 Look like you got through everything but the Sinatra part. Remember you can come back to finish this assignment if you feel you want some more practice or for review.

PR #242, opened 2017-05-30
Tim Chase Landlord w/ AR solution

Tchase44:
No Method Error on " get  '/apartments/new' " linking to new_apt.erb
but the error message is referencing apt_details.erb (which renders correctly with "get '/apartments/:num'")

tenant/resident form works, adds to the database and shows up under correct apartment.

comfort  score: 4 
difficultly score: 4 (feels like i know what to do, but just can't get it to work right)


app.rb
```diff
@@ -0,0 +1,105 @@
...
+
+get '/apartments' do
+	@apt = Apartment.all
+	erb :apt_list
+end
+
+get '/apartments/:num' do
+	@apts = Apartment.all
+	@apt_detail = []
+	@apts.each do |apt|
```
amaseda, line 38: How could you accomplish this using `.map` or `.select`?

app.rb
```diff
@@ -0,0 +1,105 @@
...
+	  			else
+	  				next
+	  			end
+	  		end
+  		end
+  	end
+  		erb :apt_with_res
+end
+
+get '/apartments/new' do
```
amaseda, line 68: Make sure this route comes before `get /aparments/:num`. Otherwise, it's going to treat `"new"` as a value for `:num`, which will cause errors.

amaseda:
:+1:

PR #243, opened 2017-05-30
Scott's landlord

scottpiercew:
Console.rb commented out text showed up after step two but not after step three.


models/tenant.rb
```diff
@@ -0,0 +1,3 @@
+class Tenant < ActiveRecord::Base
+  belongs_to :apartment
+end
```
superbuggy, line 3: Need to add views!

superbuggy:
good progress 👍

PR #245, opened 2017-05-30
Created apartments and tenants sql table. Loaded with seed file

RobLowe1992:


superbuggy:
that's a start! did you forget to push?

PR #253, opened 2017-07-26
Landlord Homework (Liz Jewell)

lizardbird:
comfort: 5
complete: 5

Everything functions as it should, but it's not a very sexy display. This essentially had the same requirements as the Pokenatra lab, but the content was far less interesting, so I did not spend any time formatting or styling this and instead spent a lot of time on Pokenatra.


AndyWhitley:
👍

PR #255, opened 2017-07-27
Alex W Landlord AR

awandres:
comfort - 4
complete - 4

Spent a lot of time trying to figure out how to do the 'menu' section. 

From the instructions, it seemed like the main menu was supposed to have all the options of what you could do with the apartments (show all, show by ID, create apartment, show tenant by apartment ID). I tried to accomplish this, but I couldn't figure out how to receive an apartment ID by input and then immediately go to the listing of that apartment via ID. Same with showing a tenant by ID (because you would have to express which apartment you wanted to view).

I may have misunderstood, but I just continued on it doing something similar to pokenatra. I didn't finish it 100% because I wanted to work on rails

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,0 +1,6 @@
+# frozen_string_literal: true
+source "https://rubygems.org"
+gem "pg"
+gem "activerecord"
+gem "pry"
+# gem "rails"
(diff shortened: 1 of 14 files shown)


juancgarcia:
👍 Great job!

So to take a second input (for instance for an ID), inside of the if selected block you would prompt again with `puts` asking for the ID and read that extra info with `gets.chomp`

PR #256, opened 2017-07-27
Charles Cox Landlord

cwcox:
completeness: 5
comfort: 4 - can transfer the sinatra CRUD functions across projects well, but need to practice more to fully understand and be able to expand this functionality.  Also need to do more practice with active record methods


juancgarcia:
👍 Great work! It looks like you're coming along strong here.

